Issue #307, opened 2014-10-24
Ignore existing documents


SHSE:
It would be great if you could tell es-hadoop to ignore existing documents when `es.write.operation` is set to `create` (e.g. `es.write.ignore.conflicts` config property).
For now it throws `EsHadoopInvalidRequest: Found unrecoverable error [Conflict(409)...`.

`update` operation requires `_source` to be enabled which is an additional and in my case unnecessary overhead.

dmarkhas:
Can this please be picked up? There are two different pull requests addressing this issue, please accept either one of them since this is breaking Elastic functionality - Elastic allows to bulk insert documents, and the operation is successful even if some already exist - the plugin should mimic this functionality, instead of throwing an exception in a perfectly valid condition (from Elastic's perspective).

jbaiera:
@dmarkhas you mention that the bulk insert operation is successful even if some documents already exist. While that is true, the response returns with a successful HTTP status, the body contains a breakdown of all errors that were encountered.

I think it would be incredibly unwise to add a blanket option to ignore conflicting writes when the entire purpose of the "create" operation in Elasticsearch is to avoid conflicting writes. Perhaps if your write patterns require ignoring existing documents and only inserting data that does not exist then instead we should talk about ways to extend the framework's handling of update/upsert operations to support that specifically.

dmarkhas:
@jbaeira, your point is valid but there is a specific use case that the current implementation breaks (which I should've mentioned earlier) - external versioning.

When using external versioning in ElasticSearch, it is perfectly acceptable to ignore conflicting writes; in fact, that is exactly what external versioning does - it provides a safeguard against overwriting documents with older versions.

We use a timestamp as the document's version to make sure that if events arrive out of order the data will still be consistent. If we already have a document that has a newer version (i.e. newer timestamp) than the one we are trying to index - we expect ES to return 409 for this document, and index the rest of the bulk successfully.

If the connector aborts the entire operation if a single document in the bulk conflicts with an existing document, that breaks our ability to use external versioning completely, and makes it impossible for us to safeguard against out of order events.

jbaiera:
Closing this in favor of #1064. Please post any conversations or feedback on the new issue.

PR #308, opened 2014-10-24
Added `es.batch.write.ignore.existing` option

SHSE:
If `es.batch.write.ignore.existing` is set to `true` then es-hadoop will ignore existing documents during a `bulk create` operation (#307).

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -75,6 +75,9 @@ public interface ConfigurationOptions {
     String ES_BATCH_WRITE_RETRY_POLICY_SIMPLE = "simple";
     String ES_BATCH_WRITE_RETRY_POLICY_DEFAULT = ES_BATCH_WRITE_RETRY_POLICY_SIMPLE;
 
+    String ES_BATCH_WRITE_IGNORE_EXISTING = "es.batch.write.ignore.existing";
+    String ES_BATCH_WRITE_IGNORE_EXISTING_DEFAULT = "false";
+
     /** HTTP connection timeout */
     String ES_HTTP_TIMEOUT = "es.http.timeout";
     String ES_HTTP_TIMEOUT_DEFAULT = "1m";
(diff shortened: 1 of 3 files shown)


aewhite:
It looks like the error condition DocumentAlreadyExistsException is hard coded in this PR. Can that be easily generalized? 

In this case it looks like you want to ignore existing documents but other use cases might include handling version conflicts and malformed mappings.

SHSE:
@costin, what would be your suggestions for the pull request?

gautamjeyaraman:
I am too having the same issue but in my case, the exception to ignore is VersionConflictEngineException caused when an older version of the same document is indexed. This PR handles only DocumentAlreadyExistsException making it too specific. Can we let the user set a list of exceptions to ignore while indexing? Can it be set in the configuration itself? Please comment on this approach so that I can send a new PR with these changes.

shawnsmith:
I'm also running into the issue where a bulk import is failing due to an ignorable version conflict exception ("Found unrecoverable error [Conflict(409) - [VersionConflictEngineException[...]]").  I'm using external version numbers where a version conflict means my batch job attempted to overwrite a more recent document with stale data.  In those cases I want to quietly ignore the failure and continue.

spoutnickgp:
Smae here, we have a complex external version number system, and we're fine with ignoring the exception gracefully (or at least quietly) in some specific cases. This feature would really come in handy.

SHSE:
If you are using Apache Spark there is native client integration lib for ElasticSearch: https://github.com/SHSE/spark-es

It allows you to "tune" each individual indexing request (version, operation type, index name, error handling, etc).

YiuTerran:
Thank you for your post. I don't know why the author have not merge this pull request. So i decide to rebuild using this patch myself.

abij:
Hi SHSE,

According to the docs and ElasticsearchException.java line 530:
// 54 was DocumentAlreadyExistsException, which is superseded by VersionConflictEngineException

Could you add check for this class and merge-in last version?

cscetbon:
@costin It would be useful to ignore non existing documents with `bulk update` too

shakdoesgithub:
Hi, 

Is there a roadmap to accept this pull request? 
We also have a need for this.  Any way I can help?

MNovak12:
Hi,

I'd like to use this feature. Any update on when it will be merged?

jbaiera:
This PR is very old, has conflicts with the main line and seems to overlap with #693. Can anyone give good use cases or reasons for ignoring errors on create or update calls that couldn't be fixed with improved upsert functionality in the connector?

jbaiera:
I am closing this PR. I have opened a new issue #1064 to discuss allowing users to decide for themselves what to do when exceptions are raised via registered error handlers.

PR #335, opened 2014-11-29
Adds a utility to walk Map[String,Any] as discussed at issues/251

jeffsteinmetz:
https://github.com/elasticsearch/elasticsearch-hadoop/issues/251

implicits available when you import:
import org.elasticsearch.spark.rdd.EsRDDUtils._

usage examples:

          val a:Double = (v._2 \ "double_test").asOpt[Double].getOrElse(0.0)
          val b:Long = (v._2 \ "long_test").asOpt[Long].getOrElse(0)
          val c:String = (v._2 \ "nested.string_test").asOpt[String].getOrElse("")
          val d:Option[String] = (v._2 \ "invalid.location").asOptString


costin:
@jeffsteinmetz As this is not needed in the Spark module, I'm thinking of closing the PR. It looks valuable so maybe it makes sense to move it into its own project or potentially contribute it to Spark?
What do you think ?

jeffsteinmetz:
feel free to close.  I'm using something similar to this in production on my own projects (started using the non json version before you added the option to allow access to original json).

j

PR #391, opened 2015-03-11
Added ability to exclude fields from the es mapping via 'es.mapping.exclude' config to ScalaValueWriter and SchemaRDDValueWriter             

rogersmarin:
This is partially referenced in issues#381; however this fix will not work when using saveJsonToEs.


costin:
Thanks! Merged in master through 9644cbdc3a805d699f63e91a83df5ab863ba418a. 
Cheers

PR #419, opened 2015-04-07
Update socks proxy configuration in  configuration.adoc

wfowlks:
fixed socks proxy network flag


costin:
Thanks. merged in master.

PR #438, opened 2015-04-29
Use javadoc of Cascading 2.6

fs111:
since you use Cascading 2.6.3 as a dependency, it should also point to the correct javadoc.


fs111:
I have signed the CLA before. Do I have to do that again?

costin:
You should not have to. Maybe you used a different email than the one in
CLA. Or maybe the verification system has an issue on our end.

fs111:
My gh account is using my private mail. Could be that i signed it w/ my company email. I will sign it again.

costin:
Danke!

fs111:
Something is foobar at your end. I signed it again and it is still red...

costin:
Sure. Will investigate - in the meantime, merging the request.

costin:
Thanks!

PR #450, opened 2015-05-14
Adding possibility to backup indexes into hdfs authorized by kerberos.

awislowski:
Adding possibility to backup indexes into hdfs authorized by kerberos.

Configuration of kerberos and hdfs can be set into elasticsearch.yml
repository:
  hdfs: 
     conf.kerberos_config: "/etc/krb5.conf"
      conf.hdfs_config: "/etc/hadoop/hdfs-site.xml"
      conf.hadoop_config: "/etc/hadoop/core-site.xml"

### DIFF
--- a/repository-hdfs/src/itest/resources/elasticsearch.yml
+++ b/repository-hdfs/src/itest/resources/elasticsearch.yml
@@ -10,5 +10,8 @@ repositories:
         concurrent_streams: "2"
         compress: "true"
         chunk_size: "1mb"
+        conf.kerberos_config: "/etc/krb5.conf"
+        conf.hdfs_config: "/etc/hadoop/hdfs-site.xml"
+        conf.hadoop_config: "/etc/hadoop/core-site.xml"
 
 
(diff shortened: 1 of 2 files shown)


awislowski:
I have already signed the CLA

costin:
@awislowski The reason the CLA test fails is because you have signed the CLA with your Github user/email but the PR relies on a different email (presumably work related).
Typically folks amend the PR with their personal email but if that's not the case, simply add your email to your github profile and the CLA will pass. Additionally, you can sign another CLA but that tends to be overkill.

Last but not least, thanks for the PR! Let me know what email you want to use so I can merge this in.

Cheers!

awislowski:
Hi Costin,
I have already addedmy email to my github profile.

Regards

costin:
@awislowski I have merged your PR in with a bit of polishing:

- remove the system wide kerberos setup; this is not needed or anything, it should not be in the module as there might be multiple instances running on the same machine.
- there's no need to add other parameters for loading configuration files - `conf_location` allows multiple values (separated through come). So the two config files in your PR can simply be added there.
- the `hadoop.*` properties are loaded from the configuration files. At least that's how it should be so there's no need to specify them manually

The resulting code is significantly smaller and also allows other configurations outside Kerberos to work as well.

Cheers,

PR #470, opened 2015-06-09
Add automatic field detection to EsHadoopScheme and EsTap.

andy327:
This feature branch overrides the retrieveSourceFields method in EsHadoopScheme. Note that it involves a REST query to extract the field names from elasticsearch, and as such, RestClient needs a public GET method.

### DIFF
--- a/cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopSearchTest.java
+++ b/cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopSearchTest.java
@@ -96,6 +96,18 @@ public class AbstractCascadingHadoopSearchTest {
     }
 
     @Test
+    public void testReadFromESWithoutFields() throws Exception {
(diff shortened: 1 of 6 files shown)


costin:
@andy327 Thanks for the PR. However the 'feature' is not part of the connector on purpose; until in its current form. The connector doesn't expect either party (Elasticsearch or Hadoop) to control the data and requires the Hadoop side to specify the mapping. That is what data it needs/expects from Elasticsearch.
For a large majority of cases, the index in Elasticsearch might be mapped entirely in Hadoop; however in others, only a subset of fields might be used.

This behaviour exists through-out all the integrations. It's also consistent - what if the index doesn't exist but is about to be created? What if the index is not uniform - the documents can vary in structure after all.
Specifying the schema, in this case the tuple might seem redundant but it really isn't; it's a small price to pay for consistency and structure.

jbaiera:
Closing this as this does not seem to have gone anywhere beyond the original discussions. Please feel free to comment if it should be re-opened.

PR #477, opened 2015-06-20
Check whether the index exists before creation

takezoe:
`RestClient` always tries to create the index at first, but Elasticsearch logs it as **ERROR** if the index already exists. It might cause an alert from monitoring system.

I guess it's better to check whether the index exists before creation to avoid error log output at Elasticsearch.


costin:
Merged in master. Thank you.

PR #503, opened 2015-07-14
Update index.adoc

mcascallares:
Typo fix


costin:
Fixed in master. Thanks for the PR!

PR #510, opened 2015-07-18
Array support for Scala serialization

takezoe:
`ScalaValueWriter` can't serialize `Array` in elasticsearch-hadoop 2.1.0.
It had been possible until 2.1.0.Beta3, so we can't upgrade to elasticsearch-hadoop 2.1.0 from Beta3.

Attached testcase success with my fix, but array cases fails with elasticsearch-hadoop 2.1.0.


costin:
@takezoe Thanks for the excellent PR. The arrays were supported through the base `JdkValueWriter` however by introducing support for JavaBeans (which is really a guess), the method call chain was bypassed. 
Thanks to your PR this has not been fixed.

Cheers,

PR #513, opened 2015-07-20
Do not route MapReduce reads and writes through non-data nodes

### LINKED ISSUE #512: Do not route MapReduce reads and writes through non-data nodes
The embedded [REST client](https://github.com/elastic/elasticsearch-hadoop/blob/a4c367b4f3f2271513f0befec3ebcc60022a11e9/mr/src/main/java/org/elasticsearch/hadoop/rest/RestClient.java) in the MapReduce integration in Elasticsearch Hadoop will by default route through [non-data nodes](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html) (e.g., dedicated master nodes). This can be problematic as non-data nodes tend to have fewer resources (e.g., network and memory) than data nodes; routing reads using [`org.elasticsearch.hadoop.mr.EsInputFormat`](https://github.com/elastic/elasticsearch-hadoop/blob/a4c367b4f3f2271513f0befec3ebcc60022a11e9/mr/src/main/java/org/elasticsearch/hadoop/mr/EsInputFormat.java) and writes using [`org.elasticsearch.hadoop.mr.EsOutputFormat`](https://github.com/elastic/elasticsearch-hadoop/blob/a4c367b4f3f2271513f0befec3ebcc60022a11e9/mr/src/main/java/org/elasticsearch/hadoop/mr/EsOutputFormat.java) through them puts pressure on these resources.

For example, dedicated master nodes usually have smaller heaps than data nodes. Reading and writing from Hadoop MapReduce jobs will cause the network buffers in the Elasticsearch JVM on these nodes to fill up leading to continual GC churn. This churn can lead to GC pauses that cause the node to be temporarily partitioned from the cluster; this is especially problematic when the node is the elected master node.

This can be avoided by disabling `es.nodes.discovery` and only specifying the data nodes in the cluster in the `es.nodes` property. However, disabling `es.nodes.discovery` and managing through `es.nodes` is not user-friendly, especially for large clusters or clusters with frequent node maintenance.

A more user-friendly approach would be to not route through non-data nodes by default but permit routing through non-data nodes via a configuration option. Since this configuration option will be enabled by default, this is a behavior change from previous versions.

jasontedor:
This commit adds a configuration option `es.nodes.data.only` for managing whether or not reads using `EsInputFormat` and writes using `EsOutputFormat` in Hadoop MapReduce jobs route through non-data nodes. The reason for wanting to avoid routing through non-data nodes is because these nodes are usually resource-constrainted relative to data nodes. This configuration option is enabled by default.

Closes #512

### DIFF
--- a/cascading/src/main/java/org/elasticsearch/hadoop/cascading/CascadingUtils.java
+++ b/cascading/src/main/java/org/elasticsearch/hadoop/cascading/CascadingUtils.java
@@ -61,6 +61,7 @@ public abstract class CascadingUtils {
 
         InitializationUtils.discoverNodesIfNeeded(settings, log);
         InitializationUtils.filterNonClientNodesIfNeeded(settings, log);
+        InitializationUtils.filterNonDataNodesIfNeeded(settings, log);
         InitializationUtils.discoverEsVersion(settings, log);
 
         InitializationUtils.setValueWriterIfNotSet(settings, CascadingValueWriter.class, log);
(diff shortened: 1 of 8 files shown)


costin:
Merged in master and 2.1.x

Issue #569, opened 2015-10-12
Support for null-safe equality comparison at Spark-Elasticsearch datasource.


HyukjinKwon:
For Spark 1.5, `EqualNullSafe` can be pushed down as a filter operator.
(https://issues.apache.org/jira/browse/SPARK-9814)

I think this can be easily done adding combination of equality comparison query and `missing` filter adding a `case` at `ElasticsearchRelation.translateFilter()` in `org.elasticsearch.spark.sql.DefaultSource.scala`

Could I open a PR for this?

costin:
I missed the new filter. Sure, go ahead with a PR (take a look at `StringStartsWith` and co to add the case in a backwards compatible way with Spark versions lower than 1.5).

Thanks,

costin:
Hi,
Thanks for the PR. It is merged in master and 2.x
I've added a null check for `EqualsTo` - for `NullSafeEquals` it doesn't seem to make sense considering nulls are handled by the operator directly (based on what I've seen in the Spark code).
The same seems to apply for String operators - in fact is null is passed, the whole filter can't really work properly since the regex becomes global (every string starts or ends with null for example).

Cheers,

HyukjinKwon:
Yeap I agree. I think this wouldn't be a big deal.
Thanks :)

PR #572, opened 2015-10-13
[SPARK] Support the filter operator of null-safe equailty comprison.

HyukjinKwon:
https://github.com/elastic/elasticsearch-hadoop/issues/569

I added the support of `EqualNullSafe`. For now it works fine and identical with `EqualTo`.

When the given value to the filter is `null`, Spark physical planning produces `IsNull` filter. So, it looks fine that it works as `EqualTo`.

However, it looks we better add some logic for null safety just in case, not only here at `EqualNullSafe` but also at `StringStartsWith`, `StringEndsWith` and `StringContains`.


HyukjinKwon:
CLA sign-up is done.

costin:
Merged in master and 2.x

Issue #592, opened 2015-11-05
esRDD Implicit with all 3 parameters (resource, query, config)


randallwhitman:
When calling `sc.esRDD` I need to specify all of resource, query, and config.  I'd like to call the implicit rather than explicitly calling `EsSpark(sc, resource, query, config)`.  In `spark/package.scala` I find:
*   `def esRDD()`
*   `def esRDD(resource: String)`
*   `def esRDD(resource: String, query: String)`
*   `def esRDD(cfg: scala.collection.Map[String, String])`
*   `def esRDD(resource: String, cfg: scala.collection.Map[String, String])`

but not a 3-parameter overload - which can be added as:
```scala
    def esRDD(resource: String, query: String, cfg: scala.collection.Map[String, String]) =
        EsSpark.esRDD(sc, resource, query, cfg)
```
```scala
    def esJsonRDD(resource: String, query: String, cfg: scala.collection.Map[String, String]) =
        EsSpark.esJsonRDD(sc, resource, query, cfg)
```

randallwhitman:
Do I understand correctly that https://github.com/elastic/elasticsearch-hadoop/commit/a97fa9fcc7ddc84478e1200338587d95cf412653 resolves this issue for v2.1.3 and v2.2.x ?

costin:
@randallwhitman That is correct. The [nightly builds](https://www.elastic.co/guide/en/elasticsearch/hadoop/master/install.html#download-dev) have been also published so you can try them out yourself.

PR #604, opened 2015-11-09
[SPARK] SparkContextFunctions.esRDD parameters

randallwhitman:
Add an overload with all the three parameters available with
EsSpark.esRDD(..., resource, query, cfg).
Likewise for esJsonRDD.

This change makes `SparkContextFunctions.esRDD` more consistent with `SQLContextFunctions.esDF` which already has an overload taking all 3 parameters.

Implements #592.


costin:
@randallwhitman I was about to push a commit fixing the issue. Can you please sign the CLA so I can use your PR instead?

Thanks,

randallwhitman:
Probably, but if you already have a commit prepared, wouldn't that be easier than processing a CLA?

costin:
The CLA should be processed automatically. As for the commit, you already went through the process of providing a fix, plus reporting the bug and I think it would be a good way to recognize your efforts by including your work in the project.
Let me know what option you prefer.

Cheers,

randallwhitman:
Thanks Costin - that makes sense, and I appreciate it.

Because I made the patch while working for my employer on integration with Elasticsearch, the CLA would have to go through some corporate process.

In this case it may be in our mutual interest to close the PR without merging, and use your independent commit.

PR #623, opened 2015-12-15
Resolve IP Address for spark.es.nodes param

pbamba:
This request is a proposition to fix the following situation : 

With the parameters : 
es.nodes.discovery = false
es.nodes.client.only = false
es.nodes.data.only = true
es.nodes = abc.example.com

With abc.example.com resolving to 1.2.3.4

The data node on abc.example is incorrectly filtered because the InitializationUtils class performs a comparison between the ip extracted from "_nodes/http" and the hostname which resolves to that ip.

When enabling debug, you then get the following situation:

"DEBUG ScalaEsRDD: Found data nodes [1.2.3.4:9200]"
"DEBUG ScalaEsRDD: Filtered discovered only nodes [abc.example.com:9200] to data-only []"

Which raises an EsHadoopIllegalArgumentException : "No data nodes with HTTP-enabled available; node discovery is disabled and none of nodes specified fits the criterion [abc.example.com:9200]"

The proposed solution resolves the ip for the hostname when qualifying nodes which allows consistency for subsequent comparisons.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/util/SettingsUtils.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/util/SettingsUtils.java
@@ -25,16 +25,19 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 
 import org.elasticsearch.hadoop.cfg.InternalConfigurationOptions;
 import org.elasticsearch.hadoop.cfg.Settings;
+import org.elasticsearch.hadoop.EsHadoopIllegalArgumentException;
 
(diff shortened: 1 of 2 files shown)


costin:
Thanks for raising this. Please sign the CLA in order to consider it.

costin:
Merged in master. Thanks for your contribution!

pbamba:
No problem @costin, thank you looking at it !

shamak:
This still doesn't work with a URL as such:

https://example.com

@pbamba any possible chance for a fix? It seems like the code is stripping out the 'https://' from the URL and then trying to resolve the hostname.

costin:
@shamak This has been fixed through #652 in master. And please, open up a new issue instead of commenting on a _close_ PR.
Thanks,

PR #673, opened 2016-01-27
Script settings

jasontedor:
This commit modifies the usage of the settings "script.inline" and
"script.indexed" as these settings only accept the values "true",
"false", and "sandbox" now.

Relates elastic/elasticsearch#16197


costin:
Merged in master through 974dd34aeb5aca08fa93b8a1d1eba3a791b74ee5
Thanks Jason!

Issue #613, opened 2015-11-19
409 is not always unrecoverable


AlexeyRaga:
Since `elasticsearch-hadoop` provides the ability to set up the external version (`es.mapping.version`) it would be nice to have a bit more control on it.

The problem is that when executing a batch that contains a version conflict the job fails with 

```
org.elasticsearch.hadoop.rest.EsHadoopInvalidRequest: Found unrecoverable error [Conflict(409) - Bailing out...
```

In some of the situations (typical dedup) it is perfectly fine to accept the fact that there could be older versions and to just ignore them.

The ES itself supports this feature: it returns status `409` for conflicting documents and `201` for the rest of the documents:

Request:

```
POST /_bulk
{ "create": { "_index": "website", "_type": "blog", "_id": "123", "_version": 10, "version_type": "external" }}
{ "title":    "My first blog post" }
{ "create": { "_index": "website", "_type": "blog", "_id": "ccc" }}
{ "title":    "My ALWAYS post" }
```

Response:

```
{
  "took": 2,
  "errors": true,
  "items": [
    {
      "create": {
        "_index": "website",
        "_type": "blog",
        "_id": "123",
        "status": 409,
        "error": {
          "type": "version_conflict_engine_exception",
          "reason": "[blog][123]: version conflict, current [10], provided [10]",
          "shard": "0",
          "index": "website"
        }
      }
    },
    {
      "create": {
        "_index": "website",
        "_type": "blog",
        "_id": "ccc",
        "_version": 1,
        "_shards": {
          "total": 2,
          "successful": 1,
          "failed": 0
        },
        "status": 201
      }
    }
  ]
}  
```

It would be great if we have the ability to have a bit more control over the conflicting situations. It should at least be possible to instruct the spark driver to not fail on `409` version conflicts.

mvervuurt:
We also have this issue, having a configuration parameter than helps ignore version conflicts with the ES-Hadoop connector especially when using external versioning would be of grate value.

jbaiera:
Closing this as it is a duplicate of #307

PR #693, opened 2016-02-07
Option to ignore version conflict errors on writes.

shawnsmith:
This is useful for external_gt and external_gte version types when write conflicts don't need to be retried or logged.  Similar to #308 but works for all conflicts, not just `DocumentAlreadyExistsException`.

relates #613

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -232,6 +232,9 @@ public interface ConfigurationOptions {
     String ES_OPERATION_DELETE = "delete";
     String ES_WRITE_OPERATION_DEFAULT = ES_OPERATION_INDEX;
 
+    String ES_WRITE_CONFLICT_IGNORE = "es.write.conflict.ignore";
+    String ES_WRITE_CONFLICT_IGNORE_DEFAULT = "false";
+
     String ES_UPDATE_RETRY_ON_CONFLICT = "es.update.retry.on.conflict";
     String ES_UPDATE_RETRY_ON_CONFLICT_DEFAULT = "0";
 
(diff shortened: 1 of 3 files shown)


dmarkhas:
I ran into the exact same situation where saveToEs throws an exception on version conflict that I want to ignore. This PR has been open for almost a year now... can it be merged?

pankaj-s:
Can we please pick this up as this is impacting multiple people and the changes are were small and already done.

alegrm:
Hello, is the solution going to be merged soon? thanks.

shirleyniara:
Hi, I am also waiting for this feature. Is there a road map to merge this branch?

jbaiera:
After having a long set of discussions internally on the right way to support this, we've now come to the conclusion that this is a feature for the connector that we're willing to support. This is an old PR, and currently has conflicts with the main branches. I'd be fine with leaving it open for a little bit, but if no progress is made I'll close this out and start fresh.

The other thing is that I'd like to try and limit the ignore functionality to only be used with version ids. I haven't seen good use cases for ignoring them in any other context, so if this PR is revived, it will need to include checks to ensure that.

shawnsmith:
I've resolved the conflicts by rebasing the branch on top of current master.

Re limiting to only be used with version ids, do you know off-hand if that data is already available to `RestClient`, perhaps in the error json returned in the `Response`?

jbaiera:
@shawnsmith It would have to be validated during the setup of the job, most likely in the InitializationUtils. Only thing I could see being a potential snag here would be that Spark handles saving data with metadata by using a different code path that doesn't set anything in the configuration.

shawnsmith:
Thanks @jbaiera, and I'm glad to hear this feature may actually get merged.  I don't think I'm going to be able to make the additional changes, though.  It would be best if someone else could take over the work.

jbaiera:
Since the original owner may not be able to pick up any additional changes here I'm going to close this out. I have opened a new issue #1064 to discuss the best way to have users to decide for themselves what to do when exceptions are raised via registered error handlers. Please post further comments on that issue, as they will have a great impact on our decisions and designs moving forward.

Issue #709, opened 2016-02-29
Heap overflow in power set creation ES 1.7.4


babsher:
I have created a test where the creation of the power set for shard sorting causes a heap overflow. 

I believe this assertion here is invalid ShardSorter.java Line 218. https://github.com/babsher/elasticsearch-hadoop/blob/master/mr/src/main/java/org/elasticsearch/hadoop/rest/ShardSorter.java#L218

See my fork https://github.com/babsher/elasticsearch-hadoop

costin:
Can be more specific on what is wrong with the aforementioned assert? It is too high which leads to the heap overflow (initially I started with 24 but as that is not that uncommon, moved to 30).

babsher:
So if you look at my test the size is exactly 30 but I still have a heap overflow. In addition this is only querying 2 indices and on our cluster. We had hoped to query at least 30 indices.

costin:
The only quick workaround I can give for the moment is to stop using aliases. The issue (as you can tell by the ShardSorter comment) is that when dealing with an alias, there's no way for the client to easily select just a certain shard that belongs only to an index without having to redo the filtering and such from the backing alias (which is just replicating ES functionality).
Potentially `ShardSorter` can be improved by using a more memory efficient `PowerSet`.

babsher:
From [RestRepository.java Line 349](https://github.com/babsher/elasticsearch-hadoop/blob/master/mr/src/main/java/org/elasticsearch/hadoop/rest/RestRepository.java#L349) it seems like any multi-index query has the shard sorter run not just aliased queries. Our query hits multiple indices and has a `,` but I am sure there are no aliases in it.

Is the `ShardSorter` also needed to query multiple indices?

costin:
Multi indices tend to have the same issue though to a lesser degree - by requesting data from all indices at once instead of separate indices, one ends up with slightly different semantics. Breaking down the multi index into smaller sub-indices leads to _subjobs_ which don't work all that well.
I'll review the issue and see whether it would be possible to perform the set lookup in a slower manner but without storing the data as a set.

babsher:
Our data is a time series log so we have no choice but to use multiple indices. For now I am testing querying one index at a time and then joining the results into a single RDD.

costin:
Which is far from ideal, I know...

babsher:
Maybe instead of generating the entire set at once the code could just iterate over the set?

babsher:
[...]

PR #721, opened 2016-03-18
[#709] Removed heap overflow from power set 

babsher:
See Issue #709


costin:
@babsher Thanks for the PR. Can you please sign the CLA? Thanks!

babsher:
I was looking over the changes. The only problem I see is that the order of the results may be different but based on the criteria for finding a good set of nodes I don't think this is an issue.

PR #751, opened 2016-04-19
Update JdkValueReader.java

yuanice:

Please make sure you have signed our [Contributor License Agreement (CLA)](https://www.elastic.co/contributor-agreement/).  
- [x] I have signed the [Contributor License Agreement (CLA)](https://www.elastic.co/contributor-agreement/)

add 'yyyy-MM-dd HH:mm:ss' support

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/serialization/builder/JdkValueReader.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/serialization/builder/JdkValueReader.java
@@ -391,13 +391,25 @@ public class JdkValueReader implements SettingsAware, ValueReader {
                 val = parseDate(parser.longValue(), richDate);
             }
             else {
-                val = parseDate(value, richDate);
+                val=getConversDate(value,richDate);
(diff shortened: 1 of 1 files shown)


costin:
@yuanice Date handling is tricky and it's also the reason why ES-Hadoop sticks to using ISO8601. If somebody wants to convert or format a date they can easily do so in the ETL layer. It's a minor inconvenience to keep doing that however it also enforces properly handling.
For example the date format above has no time zone support meaning it is unclear when it occurred.
Long term we _might_ read the time formatting support in ES and provide the same serialization in ES-Hadoop itself however that is difficult as it requires us to use the same date formatting library.
Moving to Java 8 might alleviate this concern and provide the same semantics.

jbaiera:
Closing this as it seems that:

1) CI is no longer passing
2) No further discussion has occurred after the original posts
3) The last post explains a conflict of this PR with one of the project goals.

If this should be re-openned, please say so and we can resume progress on this PR.

Issue #754, opened 2016-04-25
Why does EsBolt still use the old JAR


abtpst:
### EsBolt uses old JAR

EsBolt still uses the old jar as

`import backtype.storm.topology.IRichBolt;`

backtype.storm is not supported anymore

It should be org.apache.storm as far as i know. So, if i use the new jars for the latest storm version,i cant use the EsBolt

### Version Info

OS:         :  Win7
JVM         :  1.8
Hadoop/Storm: 1.0.0 
ES-Hadoop   :  2.3
ES          :  2.3.1

costin:
The package change in Storm 1.0.0 is quite disruptive and the reason the upgrade hasn't been made yet.
Basically by upgrading to 1.0.0, the 0.x versions stop being supported - and while it would be nice to upgrade, there are a LOT of projects that simply can't do that.
I'm not sure how to proceed to be fair; the storm folks could have still provided the old jars without making the transition so disruptive.
Maintaining two jars (storm-1x and 0.x) is too much trouble and really not the right approach.

abtpst:
thanks Costin. we can continue to use storm 0.9 as it functionally, it does not hinder us.

mazhechao:
I have sent a pr https://github.com/elastic/elasticsearch-hadoop/pull/769

costin:
Storm 1.0.x was added in master (and thus support for 0.9.x was dropped).
Fixed through #769.

Issue #743, opened 2016-04-13
Upgrade to Storm 1.0


costin:
Storm 1.0 was finally released.

sachingsachin:
The package-names of storm-constructs need to be changed perhaps ([STORM-1202](https://issues.apache.org/jira/browse/STORM-1202))

mazhechao:
I have sent a pr #769

costin:
Available in master.

PR #769, opened 2016-05-20
[#754][#743] update to compat with storm 1.x.x

mazhechao:
The most change is that the package name changes from backtype to org.apache, and just fix that.


mazhechao:
@costin  Will you have a look at it please?

costin:
Thanks @mazhechao 
As I've mentioned in #754, I'm torn on whether to upgrade to Storm 1.0 or keep 0.9 around for a while. A compromise might be to upgrade master (5.0.0) to Storm 1.0 and direct Storm users of 0.9 to 2.x

mazhechao:
I agree to upgrade master(5.0.0) to Storm 1.0, as there are many Storm users using 0.9.x now.

I notice that storm-elasticsearch has been merged into Storm project since 1.0.0. However storm-ealsticsearch doesn't support ES 2.x now. As a result, those who use storm-elasticsearch cannot use ES 2.x. If es-hadoop supports Storm 1.0, users can use both new Storm and ES at the same time.

costin:
This has been merged into master (upcoming 5.0-alpha3).
Several things to note:
1. the PR was not complete and the code did not compile (the guava classes in Storm had been moved to a different package as well).
2. Storm 1.x also depends on log4j2 instead of log4j 1 and is not really compatible. Since ES and the rest of the libraries do depend on log4j 1 resolving this was somewhat problematic. I saw you tried to resolve the conflict in the classpath but that only worked for Storm while breaking all the other libraries. However removing the conflicting slf4j api seems to be solved the issue.

Last but not least the integration tests are failing with Storm 1.x. It's the same old problem of race conditions and concurrency - starting the topology and waiting for tuple ingestion is not easy since Storm doesn't offer any utility classes (as far as I know) and while the tests do pass in isolation, in bulk things start to trip.
I've tweaked them a bit but it doesn't seem to be enough - I've raised a separate issue for this. Going forward I'm seriously thinking of moving to unit tests since each release of Storm seems to break things a bit in this department.

@mazhechao Thank for your contribution and hopefully there will be many more in the future! Cheers!

mazhechao:
@costin Thanks for your review and comment.

PR #789, opened 2016-06-17
Gradle signing plugin breaks CI workflow

jbaiera:
The `signing` plugin is enabled when a job is being built from anything other than a `SNAPSHOT` version. Travis CI does not pass when project version does not end in `SNAPSHOT`


jbaiera:
Opened the pull request so that I could test the CI without doing a direct push to master.

PR #805, opened 2016-07-15
Corrected property name

MSaifAsif:
The correct name of the property is `es.mapping.date.rich` and not `es.mapping.rich.date`


jbaiera:
Nice find! I have cherry picked this onto the 2.x branch (7d8dfc2) as well as the master branch (8c6718e). Thanks for the contribution!

PR #812, opened 2016-07-21
Add the ability to create IndexPartition based on the desired number of documents per split  

### LINKED ISSUE #778: Use new parallel reader API in ES 5.x
ES 5.x has a proposal/PR elastic/elasticsearch/pull/18237 for allowing parallel reads for an alias/index based on a query based on an arbitrary number of slices. This goes beyond the capabilities of scan/scroll in ES 1.x/2.x and it's something that is desired in highly parallel environments or those that have memory/processing constrains (but high number of instances).

jimferenczi:
This pretty big PR (sorry for the size) changes how the splits are created from an elasticsearch query.
For cluster running with version prior to v5.x:
* We create one partition for each shard of each requested index.
* If an alias is requested the search routing and the alias filter are respected.
* The partition is no longer attached to a node nor an ip. Only the shardId and index name are defined in order to be able to use any replica in the cluster when the partition is consumed. This makes the retry possible if a node disapears during a job.
* The ability to consume a partition on the node that is responsible for the index/shardId has been removed temporarily and should be re-added in a follow up.

For cluster ruuning with version v5.x:
* We first split by index then by shard and finally by the maximum number of documents allowed per partition (configurable through the new option named es.input.maxdocsperpartition. For instance an index with 5 shards, 1M documents and a maximum number of documents allowed per partition equals to 100,000, a match all query would be splitted in 50 partitions, 10 partitions per shard.
* If an alias is requested the search routing and the alias filter are respected.

Fixes #778


jimferenczi:
@costin @jbaiera, this PR introduces the SlicePartitionDefinition. I had to refactor (cleanup) some parts to make it work. This is still a work in progress and I need to add more tests. In the meanwhile it would be very helpful if you can check the progress. Considering the size of the PR I am available at your convenience for a deep dive.

mr/src/main/java/org/elasticsearch/hadoop/rest/RestClient.java
```diff
@@ -410,7 +408,7 @@ public void refresh(Resource resource) {
         sb.setLength(sb.length() - 1);
         sb.append("\n]}");
 
-        if (!isES20) {
+        if (esVersion().equals(EsVersion.V_2_X)) {
```
jbaiera: Should this be `!esVersion().equals(EsVersion.V_2_X)`?

jbaiera: Actually, This should just be if the version is equal to `V_1_X`. This line is balancing the curly braces.

jbaiera:
This _is_ a very large PR. I'll be leaving comments as I find items.

mr/src/main/java/org/elasticsearch/hadoop/util/SettingsUtils.java
```diff
@@ -162,32 +162,6 @@ public static void setFilters(Settings settings, String... filters) {
...
-        // assume ES 2.0 by default
-        if (!StringUtils.hasText(version)) {
-            return true;
-        }
-
-        return version.startsWith("2.");
-    }
-    
     /**
      * Whether the settings indicate a ES 5.0.x (which introduces breaking changes) or otherwise.
```
jbaiera: Are you going to remove this settings check as well in favor of the new route? I'm assuming this is an item that is still in progress...

mr/src/main/java/org/elasticsearch/hadoop/util/EsVersion.java
```diff
@@ -0,0 +1,77 @@
+package org.elasticsearch.hadoop.util;
+
+import org.elasticsearch.hadoop.EsHadoopIllegalArgumentException;
+
+public class EsVersion {
+    public static final EsVersion V_0_X = new EsVersion(0, "0.x");
+    public static final EsVersion V_1_X = new EsVersion(1, "1.x");
+    public static final EsVersion V_2_X = new EsVersion(2, "2.x");
+    public static final EsVersion V_5_X = new EsVersion(3, "3.x");
```
jbaiera: I love the addition of this class. One nit: Can we name the `V_5_X` raw name to be `"5.x"` instead of `"3.x"`?

jbaiera: After thinking about it a little bit. Could you also set the integer value to `5` to match the raw version string as well? It's the nitty-est of picks, but consistency is nice.

mr/src/main/java/org/elasticsearch/hadoop/serialization/dto/IndicesAliases.java
```diff
@@ -0,0 +1,96 @@
...
+ * under the License.
+ */
+package org.elasticsearch.hadoop.serialization.dto;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class IndicesAliases {
+    private Map<String, Map<String, Alias> > indices;
```
jbaiera: Could be made `final`.

mr/src/main/java/org/elasticsearch/hadoop/serialization/dto/IndicesAliases.java
```diff
@@ -0,0 +1,96 @@
...
+
+                if (metadata.containsKey("search_routing")) {
+                    searchRouting = (String) metadata.get("search_routing");
+                }
+                if (metadata.containsKey("index_routing")) {
+                    indexRouting = (String) metadata.get("index_routing");
+                }
+                if (metadata.containsKey("filter")) {
+                    filter = (Map<String, Object>) metadata.get("filter");
+                }
```
jbaiera: Should these be `((Map<String, Object>)value).containsKey("...")`? The `metadata` map is only ever going to have one value named `aliases`.

mr/src/main/java/org/elasticsearch/hadoop/rest/query/TermQueryBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public TermQueryBuilder field(String field) {
+        this.field = field;
+        return this;
+    }
+
+    public TermQueryBuilder term(String term) {
+        this.term = term;
+        return this;
+    }
```
jbaiera: Is there ever a point in time where it makes sense for either of these values to be left uninitialized? I think that it should throw an illegal state exception if it's called without them being set. Better yet, maybe it makes sense to force the user to acknowledge the properties as missing by making both fields part of the constructor/static factory method?

costin:
Jim, thanks again for starting to work on this.
As you mentioned, this is quite a big PR which makes merging and reviewing tricky. Breaking it down into several smaller pieces would definitely help.

A couple of things from the top of my head:

A. Regarding the supported ES versions - currently ES-Hadoop supports 1.x, 2.x and master adds support for 5.x. Going forward, remains to be decided whether support for 1.x will be kept - either way I would not bother with 0.X

B. Also things like mapping are resolved eagerly to avoid having each client/task do that - basically for N tasks, currently there's only one mapping fetch (which also does some validation); if it's lazy it translates to N mapping calls.

Fwiw, I favor keeping the new slice feature in a separate package and try to keep the old code around to make the transition and code porting between 2.x and 5.x branches easy.

jimferenczi:
Thanks @costin and @jbaiera for the first round of review.

> As you mentioned, this is quite a big PR which makes merging and reviewing tricky. Breaking it down into several smaller pieces would definitely help.

I splitted the big commit in smaller pieces. I've intentionally kept the error spotted by your review in the original commits. The commits that address your comments are last. I hope this can help the review process.

> Also things like mapping are resolved eagerly to avoid having each client/task do that - basically for N tasks, currently there's only one mapping fetch (which also does some validation); if it's lazy it translates to N mapping calls.

I don’t see it that way, IMO keeping the mapping (and the config) in each PartitionDefinition is worst than having to retrieve it lazily when the task is initialized. The mapping (and the config) can be big and it's even worst with the base64 conversion, furthermore if the mapping is resolved when the partitions are created there is a high probability that a mapping update happens between the initialization of the job and the execution of the task. 

> Fwiw, I favor keeping the new slice feature in a separate package and try to keep the old code around to make the transition and code porting between 2.x and 5.x branches easy.

There should be no code porting between 2.x and 5.x. There is no breaking change for the available clients in the repo. The changes are internal, the only thing that client should change is the new configuration option to set the expected number of documents per task. 
The changes on this PR are not all about the slice feature. For cluster that runs on versions prior to 5 the removal of the ShardSorter (which is possible now that we check for filter and routing aliases) and the fact that each partition is assigned to an index shard (and not a node and a shard id) is already a big win. Not to mention that the removal of the mapping and the config in each partition definition should have a positive impact on the required memory to run a big job.

mr/src/main/java/org/elasticsearch/hadoop/rest/RestClient.java
```diff
@@ -588,12 +602,15 @@ public void putMapping(String index, String mapping, byte[] bytes) {
         execute(PUT, mapping, new BytesArray(bytes));
     }
 
-    public String esVersion() {
-        Map<String, String> version = get("", "version");
-        if (version == null || !StringUtils.hasText(version.get("number"))) {
-            return "Unknown";
+    public synchronized EsVersion esVersion() {
```
jbaiera: I think this should be reverted back to being retrieved from the configuration at construction time. This method is called many times in the client and the synchronized nature of it is not great for performance. There really shouldn't be a scenario where the value of the version will be different between tasks, so it's reasonable for the version information to be stored once for the entire job.

jimferenczi:
> This is removing the ability for Hadoop to attempt to schedule days local tasks. I'm not sure if this is a big deal though since most people run elasticsearch on separate hardware from their running tasks.

Right, we discussed this with @costin and he said approximatively the same so I removed it. 
If it's important we could check the list of nodes that could answer the query and pick the local one if exists. But yeah not sure it's really helpful and could be counterproductive.
I think that the nice part of this is that a task can run on any node now which makes retries of the same task possible even if some nodes are not available on the cluster.

jimferenczi:
@jbaiera about the Elasticsearch version discovery in the RestClient I pushed https://github.com/elastic/elasticsearch-hadoop/pull/812/commits/f101370f7b2ba540e5cfc8da4adc81feb935107c

This change retrieves an internal version from the settings and use it to build the requests.
Since the settings are not part of the PartitionDefinition anymore the elasticsearch version is retrieved and added in the settings each time a Partition{Reader/Writer} is built.
I think it's a good compromise, we don't have the cost of the duplicate serialized settings in each partition but we have to make a light request once per task/split to get the version.

jbaiera:
Re: Data Locality - If we're all +1 on removing support for the locality features then the docs should probably be updated to remove the mentions to data locality as well.

jimferenczi:
@jbaiera @costin I've updated the description of the PR with the latest status.
It seems that the integration tests are all green but we'll need to do some backward compatibility tests at least with elasticsearch v2.x.
I did not restore the data locality yet. It's trickier than I thought so I wonder if it would be ok to merge this PR without this functionality for alpha5 and make it a blocker for 5.0 ?

mr/src/main/java/org/elasticsearch/hadoop/rest/query/RawQueryBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+    private final String queryString;
+
+    public RawQueryBuilder(String value) throws IOException {
+        this((Map<String, Object>) MAPPER.readValue(value, HashMap.class));
+    }
+
+    public RawQueryBuilder(Map<String, Object> map) throws IOException {
+        Object query = map;
+        if (map.containsKey("query")) {
+            query = map.remove("query");
```
jbaiera: This line right here breaks backwards compatibility with ES v1.7.3 for SparkSQL. We have code that hooks into Spark that allows us to make sense of some of the SQL operations being executed and translate those operations into pushdown filters and queries. 

In the case of running against ES 1.7.3, the framework will detect the version and will create a query context with the pushdown query underneath it to allow compatibility with the filter context. Because all the filters are passed through the parsing code listed above before building the final query, the query context marker is removed from the JSON and the query ceases to compile from within the filter context.

Basically, the filter fragment goes from being this:
```
{ "query": { "match" : { "..." : "..." } } }
```
to this:
```
{ "match": { "..." : "..." } }
```
which is then inserted into a filter context and subsequently comes back from the server with the following error: 
```
org.elasticsearch.search.SearchParseException: [defaultstrictspark-test][2]: from[-1],size[-1]: Parse Failure [Failed to parse source [{"query":{"filtered":{"query":{"match_all":{}},"filter":{"match":{"airport":"OTP"}}}}}]]
	at org.elasticsearch.search.SearchService.parseSource(SearchService.java:747)
	<...>
Caused by: org.elasticsearch.index.query.QueryParsingException: [defaultstrictspark-test] No filter registered for [match]
```

This probably isn't a huge blocker for the alpha release, but it is a big deal if we want to continue the support for 1.x.

jbaiera:
Local integration tests are looking good for 5.x and 2.x Elasticsearch versions. Ran into some snags with 1.x surrounding SparkSQL that I have commented on above.

jimferenczi:
Good catch @jbaiera. I pushed a fix to handle 1.x query DSL.

Regarding the data locality I don't think that it works as expected in the current version (prior to this PR). The preference API is used to restrict the query to the node where the targeted shard is present but the rest client will use a random node to send the query which ultimately will be redirected to the targeted node. IMO this extra round trip makes the data locality useless. The task can run on the same machine than the node which hosts the shard but since we use a random node to do the coordination there is no advantage of doing this.

jimferenczi:
Just discoverer that SettingsUtils.pinNode is supposed to restrict the client to a node. This means that my last comment about data locality is wrong ;).
I'll continue to dig on how we could have the data locality back without breaking the fact that a partition can now run on any replica of the shard.

jimferenczi:
I pushed a fix for the NodeInfo parsing bug:
https://github.com/elastic/elasticsearch-hadoop/pull/812/commits/e6f046bb5075fbd34c32f74891135b7ee9d663c7

This fixes the discovery and the selection of nodes based on some criteria (isClientOnly, isDataOnly, ...). Please note that the bug is not due to this PR, the reason is that the format of the node info changed in 5.x.
@costin can you take a look at this patch ?

jbaiera:
re: Node Info - LGTM
Thanks for taking a look at the 1.x compatibility issues as well.

jbaiera:
I have performed the manual merge steps to get this PR into master. The commits on master can be found at 77f2564 (main pr) and 2e53b63 (node info fix). Thanks so much @jimferenczi!

PR #819, opened 2016-08-03
Restores the ability to prefer local nodes when creating a PartitionReader

### LINKED ISSUE #814: Restore data locality preference for hadoop and spark
The data locality preference has been dropped in https://github.com/elastic/elasticsearch-hadoop/pull/812. We should restore the functionality but instead of setting one node per partition we should let hadoop/spark choose among the nodes that host the index/shard targeted by the partition. Since it's only a preference, having multiple hosts that could serve the query is beneficial in terms of execution. The list of hosts should be shuffled in order to spread the execution of multiple partitions that target the same index/shard.

jimferenczi:
In Hadoop and Spark it is possible to set the list of hostnames/ips that should be preferred when a task is launched.
This change restores the heuristic that prefers ES nodes that are local to the task when a partition reader is created.
This means that if an ES node is on the same machine than an Hadoop/Spark worker and this node contains the targeted index/shardId then the PartitionReader will request this node in priority.

Closes #814

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/mr/EsInputFormat.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/mr/EsInputFormat.java
@@ -81,8 +81,7 @@ public class EsInputFormat<K, V> extends InputFormat<K, V> implements org.apache
 
         @Override
         public String[] getLocations() {
-            // TODO: check whether the host name needs to be used instead
-            return new String[] {};
+            return partition.getHostNames();
         }
 
         @Override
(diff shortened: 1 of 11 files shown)


jbaiera:
PR looks great @jimferenczi! Thanks!

chernals:
I'm confused after reading 373 and 812  : do we support data locality with Spark at the moment ?

jbaiera:
@chernals #812 broke the data locality features, but this PR brought them back into working order.

chernals:
Thanks! More as a general question (and I know this is not the right place): can you comment/refer to something somewhere I can read on how the situation of ES+Spark compares to Cassandra+Spark in term of data locality?

jbaiera:
That would be something more suited to our forums: https://discuss.elastic.co/c/elasticsearch-and-hadoop

PR #823, opened 2016-08-09
fix memory leak caused by re-wrapping DelegatedProtocol

jaymode:
When SSL is used to communicate with elasticsearch, we replace the Protocol for
`https` with a custom protocol that uses a custom created SSLSocketFactory. The
DelegatedProtocol we create wraps the existing protocol, which is fine if we only
wrap a single time. However if the http transport is instantiated multiple times,
the protocol that is wrapped will be the DelegatedProtocol that we created previously
and this causes the existing SSLSocketFactory to be retained indefinitely.

This change adds a check to see if the existing Protocol is a DelegatedProtocol and
if so the original Protocol is retrieved so that we do not retain a reference to the
DelegatedProtocol and its SSLSocketFactory.


jbaiera:
LGTM

PR #826, opened 2016-08-19
Explicitly specify `buildScan` return type

pfcoperez:
Because of type inference, not explicitly providing providing a return type for:

`def buildScan(requiredColumns: Array[String], filters: Array[Filter])`

... gives it the return type `ScalaEsRowRDD` which is a compatible type with the overridden method but limits further overriding on which some transformation is performed by the means of Spark. e.g:

```scala
org.evilcorpt.spark.sql

class SomeNewClass(...) extends ElasticsearchRelation(...) {
...
    def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] =
        super.buildScan(requiredColumns, filters).map(someTransformationFunction)
...
}
```


jbaiera:
Everything looks good to me. Thanks for the PR!

pfcoperez:
@jbaiera Thanks to you for merging it :smile:

PR #831, opened 2016-08-24
Fix NPE when es.nodes.discovery is set to false

giurim:
JavaEsSpark failed to initialise when we set discovery to false and throw an NPE
```
new SparkConf()
    .setAppName("test").setMaster("local[8]")
    .set("es.nodes", "localhost")
    .set("es.port", "9200")
    .set("es.nodes.discovery","false")
    .set("es.nodes.wan.only","true")
```
It was only working with discovery on:

```
new SparkConf()
    .setAppName("test").setMaster("local[8]")
    .set("es.nodes", "localhost")
    .set("es.port", "9200")
    .set("es.nodes.discovery","true")
    .set("es.nodes.wan.only","false")
```

This PR provides a workaround and populates the `nodes` list with the help of the client instead of discovery if discovery is off.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java
@@ -256,11 +256,10 @@ public abstract class RestService implements Serializable {
                 }
             }
             final Map<String, NodeInfo> nodesMap = new HashMap<String, NodeInfo>();
-            if (nodes == null){
-                nodes = client.getRestClient().getHttpNodes(false);
-            }
-            for (NodeInfo node : nodes) {
-                nodesMap.put(node.getId(), node);
+            if (nodes != null) {
+                for (NodeInfo node : nodes) {
+                    nodesMap.put(node.getId(), node);
+                }
             }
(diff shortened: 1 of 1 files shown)


mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java
```diff
@@ -256,6 +256,9 @@ public void remove() {
                 }
             }
             final Map<String, NodeInfo> nodesMap = new HashMap<String, NodeInfo>();
+            if (nodes == null){
+                nodes = client.getRestClient().getHttpNodes(false);
```
jbaiera: This method *is* the discovery process. We shouldn't be discovering nodes as a work around to null pointers on disabled discovery. I think a better option might be to simply return an empty list on line 218 instead of null.

jbaiera:
Thanks for opening a PR for this! I left a comment inline with the commit.

giurim:
Great,  thx.

giurim:
Removed the alternative "discovery". Now we only prevent the NPE when there is no discovery.

jbaiera:
LGTM, thanks!

PR #864, opened 2016-10-10
Throw clean exception if query is not well formed

ebuildy:
- [x] I have signed the [Contributor License Agreement (CLA)][]

[Contributor License Agreement (CLA)]: https://www.elastic.co/contributor-agreement/

To avoid "NullPointerException" if query is not a valid ressource, could help people to understand whats happened.

PR #866, opened 2016-10-10
Throw clean exception if query is not well formed

ebuildy:

As discussed at https://github.com/elastic/elasticsearch-hadoop/pull/864#issuecomment-252745629, to avoid "NullPointerException" if query is not a valid ressource, could help people to understand whats really happened.


jbaiera:
+1 Thanks!

Issue #874, opened 2016-10-20
shard preference concatenation with | gives query error


megri:
### What kind an issue is this?

- [x] Bug report. If you’ve found a bug, please provide a code snippet or test to reproduce it below.   
    The easier it is to track down the bug, the faster it is solved. 
- [ ] Feature Request. Start by telling us what problem you’re trying to solve.  
    Often a solution already exists! Don’t send pull requests to implement new features without 
    first getting our support. Sometimes we leave features out on purpose to keep the project small.

### Issue description

[This change](https://github.com/elastic/elasticsearch-hadoop/commit/fdc7d11dff5cac6ac03406f492dd3f68059ee57f#diff-e9a55df82f2607661a3937ea833e9eacL196
) results in 

```json
{"error":{"root_cause":[{"type":"number_format_exception","reason":"For input string: \"0|_local\""}],"type":"number_format_exception","reason":"For input string: \"0|_local\""},"status":400}%
```

### Steps to reproduce

Launch spark-shell with `./bin/spark-shell --packages --packages org.elasticsearch:elasticsearch-spark-20_2.11:5.0.0-rc1`

**NOTE**
Running with `org.elasticsearch:elasticsearch-spark-20_2.11:5.0.0-beta1` works

Code:

```
scala> import org.elasticsearch.spark._
import org.elasticsearch.spark._

scala> val rdd = sc.esRDD("myindex")
rdd: org.apache.spark.rdd.RDD[(String, scala.collection.Map[String,AnyRef])] = ScalaEsRDD[0] at RDD at AbstractEsRDD.scala:19

scala> rdd.count
```

Strack trace:

```
➜  spark-2.0.1-bin-hadoop2.7 ./bin/spark-shell --conf spark.es.nodes.wan.only=true --packages org.elasticsearch:elasticsearch-spark-20_2.11:5.0.0-rc1
Ivy Default Cache set to: /***/.ivy2/cache
The jars for the packages stored in: /***/.ivy2/jars
:: loading settings :: url = jar:file:/***/dev/labs/spark-2.0.1-bin-hadoop2.7/jars/ivy-2.4.0.jar!/org/apache/ivy/core/settings/ivysettings.xml
org.elasticsearch#elasticsearch-spark-20_2.11 added as a dependency
:: resolving dependencies :: org.apache.spark#spark-submit-parent;1.0
    confs: [default]
[...]

PR #877, opened 2016-10-20
encodes : and | when using them in url

bwf93:

- [ X] I have signed the [Contributor License Agreement (CLA)][]

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
@@ -187,12 +187,12 @@ public class SearchRequestBuilder {
 
         StringBuilder pref = new StringBuilder();
         if (StringUtils.hasText(shard)) {
-            pref.append("_shards:");
+            pref.append("_shards%3A");
             pref.append(shard);
         }
         if (local) {
             if (pref.length() > 0) {
-                pref.append("|");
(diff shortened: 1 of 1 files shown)


jbaiera:
I think the CLA is just a little slow on the uptake. Hopefully it will refresh.

jbaiera:
Looks like it hasn't updated yet. I'll see what's going on.

megri:
I don't think this should be merged. Please see comment on #874

jbaiera:
@megri The issue that this PR is solving is a `org.apache.commons.httpclient.URIException: Invalid query`, which is thrown during the URI parsing before the request is even sent to Elasticsearch. I am thinking this has to do with an unsupported version of commons httpclient. That being said, I think that this is still a win if it means that a simple encoding change will allow a different client version to just work, even if it's not officially supported.

karmi:
Hi @bwf93 , we have your CLA signature on file, but with a different e-mail. Can you please add both to your Github profile (they can be hidden) so we can correctly match your signature to your Github account and the Git commit?

jbaiera:
Due to time constraints surrounding backwards compatibility testing, I elected to implement a fix for #874 in commit https://github.com/elastic/elasticsearch-hadoop/commit/d442bb4479c2ae8478a10f25a7f0edb6b7256d87. The commit should catch the issues referenced in this PR. Because of this, the PR is no longer needed, and will be closed.

Issue #871, opened 2016-10-17
Pig fails to return data when `es.output.json` is enabled.


jbaiera:
When EsStorage reads data [it automatically assumes that it will receive a Map value back from the reader](https://github.com/elastic/elasticsearch-hadoop/blob/77f2564324b88a54ccf0c28736b1e8bb696fa21f/pig/src/main/java/org/elasticsearch/hadoop/pig/EsStorage.java#L263-L272). When `es.output.json` is enabled, the record reader implementation switches to be the PigJsonEsInputRecordReader which returns a `String` value.

The value returned from the ScrollReader when `es.output.json` is enabled is of type `String`. This value needs to be accepted into the RecordReader implementation so that it may transfer the data along to the users of the record reader, in this case, EsStorage. Either the RecordReader method signature should be made broader than it currently is to allow for it to accept a String value and wrap it as a Map object on retrieval, or the EsStorage object must be modified to expect anything from a record reader, and should it be an implementation of anything other than a Map, to implicitly wrap it into a singleton Map. I'm in favor of the second option for the sake of type consistency in the RecordReaders.

jbaiera:
Fixed in 5.1.1.

PR #897, opened 2016-11-17
Update to EsStorage to account for JSON output mode

jbaiera:
If `es.output.json` is enabled, instead of deserializing JSON data into a map object ES-Hadoop will simply return the raw JSON string to the user. This is returned from the Hadoop RecordReader as a `Text` object containing the string data. When the Pig integration receives data from a RecordReader, it automatically assumes that the data is a Map object. This leads to a cast exception.

This bug fell through for a while because the tests surrounding reading JSON data only check to see if the previously inserted raw JSON data can be read back out traditionally. I have added a new test that will make sure that same data can be read back out as raw JSON.

This relates to the issue #871.


jbaiera:
Merging this in as it's a pretty small change.

Issue #885, opened 2016-10-27
Cascading fails to return data when `es.output.json` is enabled.


jbaiera:
When EsHadoopScheme receives data from it's configured record readers (hadoop case) or scroll queries (local case) it automatically assumes that it will receive a Map value. When es.output.json is enabled, the record reader implementation switches to be the JsonWritableEsInputRecordReader which returns a Text value, and in the case of local mode enabled, the value returned from the scroll query used is a String value.

jbaiera:
Fixed in 5.1.1

PR #898, opened 2016-11-17
Fix `es.output.json` for Cascading.

jbaiera:
If `es.output.json` is enabled, instead of deserializing JSON data into a map object ES-Hadoop will simply return the raw JSON string to the user. This is returned from the Hadoop RecordReader as a `Text` object containing the string data. When the Cascading integration receives data from a RecordReader or from a ScrollQuery, it automatically assumes that the data is a Map object. This leads to a cast exception when it receives a Text object instead.

This relates to issue #885.

### DIFF
--- a/cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopJsonReadTest.java
+++ b/cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopJsonReadTest.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to Elasticsearch under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.elasticsearch.hadoop.integration.cascading;
+
+import cascading.flow.hadoop.HadoopFlowConnector;
+import cascading.operation.AssertionLevel;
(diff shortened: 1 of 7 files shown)


cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopJsonReadTest.java
```diff
@@ -0,0 +1,97 @@
...
+        RestUtils.refresh(indexPrefix + "cascading-hadoop");
+    }
+
+    @Test
+    public void testReadFromES() throws Exception {
+        Tap in = new EsTap(indexPrefix + "cascading-hadoop/artists");
+        Pipe pipe = new Pipe("copy");
+
+        Tap out = new HadoopPrintStreamTap(Stream.NULL);
+        build(cfg(), in, out, pipe);
```
tlrx: For my education, do we check the result somewhere?

cascading/src/main/java/org/elasticsearch/hadoop/cascading/EsHadoopScheme.java
```diff
@@ -83,12 +80,13 @@
...
         super.sourcePrepare(flowProcess, sourceCall);
 
-        Object[] context = new Object[3];
+        Object[] context = new Object[4];
         context[0] = sourceCall.getInput().createKey();
         context[1] = sourceCall.getInput().createValue();
         // as the tuple _might_ vary (some objects might be missing), we use a map rather then a collection
         Settings settings = loadSettings(flowProcess.getConfigCopy(), true);
         context[2] = CascadingUtils.alias(settings);
+        context[3] = settings.getOutputAsJson();
```
tlrx: Would it make sense to use a constant for the index in the context array, something like 
` context[OUTPUT_AS_JSON] = settings.getOutputAsJson();`

That would help to read the context array as well as quickly identify where it is read & write.

cascading/src/itest/java/org/elasticsearch/hadoop/integration/cascading/AbstractCascadingHadoopJsonReadTest.java
jbaiera: There's no great way to check the result in this test case since it's writing in a separate process. We're mostly just checking that the job completes instead of throwing.

cascading/src/main/java/org/elasticsearch/hadoop/cascading/EsHadoopScheme.java
jbaiera: I pushed 9ef7ffe4739128e9ddaf14239f42f0dada4cde4c

Issue #896, opened 2016-11-16
Hive fails to return data when `es.output.json` is enabled.


jbaiera:
When the EsSerde object used for the Hive integration receives data to be deserialized, it automatically assumes that the data is a Map object. In the case that `es.output.json` is set to `true`, the data that is received is a Text object that contains the raw JSON data. This causes a class cast exception during the execution.

jbaiera:
Fixed in 5.1.1

PR #899, opened 2016-11-17
Fix `es.output.json` for Hive

jbaiera:
If `es.output.json` is enabled, instead of deserializing JSON data into a map object ES-Hadoop will simply return the raw JSON string to the user. This is returned from the Hadoop RecordReader as a `Text` object containing the string data. When the Hive integration receives data from a RecordReader, it automatically assumes that the data is a Map object. This leads to a cast exception when it receives a Text object instead.

In order for Hive to correctly accept the data, it must conform structurally to Hive's table schema. Instead of requiring the user to name their string field a certain way, the code now checks if it is expecting raw JSON data and selects an appropriate schema field from the table definition to place the information into. All other fields will be left as nulls.

This relates to issue #896.


jbaiera:
Merged as it's been open for 20 days and it's not a huge change.

Issue #924, opened 2017-01-30
scala.MatchError on reading metadata


drudim:
### What kind an issue is this?

- [x] Bug report.

### Issue description

Getting `scala.MatchError` while trying to read `_metadata`.

### Steps to reproduce

Code:
```
# Fails
scala> spark.read.format("org.elasticsearch.spark.sql").option("es.read.metadata", "true").option("es.nodes", "xxx").load("yyy/zzz").select("_metadata")

# Fails 
scala> spark.read.format("org.elasticsearch.spark.sql").option("es.read.metadata", "true").option("es.nodes", "xxx").load("yyy/zzz").select("_metadata", "some_field")

# Works!
scala> spark.read.format("org.elasticsearch.spark.sql").option("es.read.metadata", "true").option("es.nodes", "xxx").load("yyy/zzz").select("some_field", "_metadata")
```

Strack trace:
```
17/01/30 21:54:51 ERROR TaskSetManager: Task 0 in stage 2.0 failed 1 times; aborting job
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 2.0 failed 1 times, most recent failure: Lost task 0.0 in stage 2.0 (TID 2, localhost): scala.MatchError: true (of class java.lang.Boolean)
       	at org.apache.spark.sql.catalyst.CatalystTypeConverters$MapConverter.toCatalystImpl(CatalystTypeConverters.scala:202)
       	at org.apache.spark.sql.catalyst.CatalystTypeConverters$MapConverter.toCatalystImpl(CatalystTypeConverters.scala:194)
       	at org.apache.spark.sql.catalyst.CatalystTypeConverters$CatalystTypeConverter.toCatalyst(CatalystTypeConverters.scala:103)
       	at org.apache.spark.sql.catalyst.CatalystTypeConverters$$anonfun$createToCatalystConverter$2.apply(CatalystTypeConverters.scala:403)
       	at org.apache.spark.sql.execution.RDDConversions$$anonfun$rowToRowRdd$1$$anonfun$apply$2.apply(ExistingRDD.scala:67)
       	at org.apache.spark.sql.execution.RDDConversions$$anonfun$rowToRowRdd$1$$anonfun$apply$2.apply(ExistingRDD.scala:64)
       	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
[...]

PR #930, opened 2017-02-11
Fix spark sql queries with metadata field

drudim:
The PR solves `scala.MatchError` in Spark SQL queries with `_metadata` field inside.
More details on the issue: https://github.com/elastic/elasticsearch-hadoop/issues/924

I was trying to update integration tests with that case,
but I am getting failed tests even for the current code in branch `5.1`.

[x] I have signed the [Contributor License Agreement (CLA)][]

### DIFF
--- a/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
+++ b/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
@@ -277,17 +277,16 @@ private[sql] object SchemaUtils {
     val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
     val sourceFields = SettingsUtils.determineSourceFields(settings)
 
-    // In case when user selected specific fields (we want to keep the same order in a spark sql row)
-    if (StringUtils.hasText(requiredFields))
+    if (StringUtils.hasText(requiredFields)) {
+      // In case when user selected specific fields (we want to keep the same order in a spark sql row)
       rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
-
-    // In case when we read all fields including metadata
-    else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata)
+    } else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata) {
+      // In case when we read all fields including metadata
       rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
-
-    // In case when we read all fields without metadata
-    else if (StringUtils.hasText(sourceFields))
+    } else if (StringUtils.hasText(sourceFields)) {
+      // In case when we read all fields without metadata
       rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields)
+    }
     rowInfo
   }
 
--- a/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
+++ b/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
@@ -273,16 +273,21 @@ private[sql] object SchemaUtils {
     val rowInfo = (new Properties, new Properties)
 
     doDetectInfo(rowInfo, ROOT_LEVEL_NAME, struct)
-    val csv = SettingsUtils.determineSourceFields(settings)
-    // if a projection is applied (filtering or projection) use that instead
-    if (StringUtils.hasText(csv)) {
-      if (settings.getReadMetadata) {
-        rowInfo._1.setProperty(ROOT_LEVEL_NAME, csv + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
-      }
-      else {
-        rowInfo._1.setProperty(ROOT_LEVEL_NAME, csv)
-      }
-    }
+
+    val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
+    val sourceFields = SettingsUtils.determineSourceFields(settings)
+
+    // In case when user selected specific fields (we want to keep the same order in a spark sql row)
+    if (StringUtils.hasText(requiredFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
+
+    // In case when we read all fields including metadata
+    else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata)
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
+
+    // In case when we read all fields without metadata
+    else if (StringUtils.hasText(sourceFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields)
     rowInfo
   }
 
--- a/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
+++ b/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
@@ -271,22 +271,13 @@ private[sql] object SchemaUtils {
   def detectRowInfo(settings: Settings, struct: StructType): (Properties, Properties) = {
     // tuple - 1 = columns (in simple names) for each row, 2 - what fields (in absolute names) are arrays
     val rowInfo = (new Properties, new Properties)
-
     doDetectInfo(rowInfo, ROOT_LEVEL_NAME, struct)
 
     val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
-    val sourceFields = SettingsUtils.determineSourceFields(settings)
-
     if (StringUtils.hasText(requiredFields)) {
-      // In case when user selected specific fields (we want to keep the same order in a spark sql row)
       rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
-    } else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata) {
-      // In case when we read all fields including metadata
-      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
-    } else if (StringUtils.hasText(sourceFields)) {
-      // In case when we read all fields without metadata
-      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields)
     }
+
     rowInfo
   }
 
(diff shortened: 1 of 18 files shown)


jbaiera:
@drudim Sorry for the very delayed reviews. I'm taking a look through this right now and it looks like it's written against 5.1. Could you rebase these changes on top of 5.x? There are a few changes at the moment upstream that have to do with source field filtering that you may want to get pulled into this PR.

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/Utils.java
```diff
@@ -44,6 +44,9 @@ static FieldType extractType(Field field) {
     // double filtering (run Spark filters) or not
     static final String DATA_SOURCE_KEEP_HANDLED_FILTERS = "es.internal.spark.sql.pushdown.keep.handled.filters";
 
+    // columns selected by Spark SQL query
+    static final String DATA_SOURCE_REQUIRED_COLUMNS = "es.internal.spark.sql.required.columns";
```
jbaiera: Could we change this to a different location? Like `InternalConfigurationOptions`?

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
```diff
@@ -271,16 +270,22 @@ private[sql] object SchemaUtils {
...
-      if (settings.getReadMetadata) {
-        rowInfo._1.setProperty(ROOT_LEVEL_NAME, csv + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
-      }
-      else {
-        rowInfo._1.setProperty(ROOT_LEVEL_NAME, csv)
-      }
-    }
+
+    val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
+    val scrollFields = settings.getScrollFields()
```
jbaiera: This logic has changed a little bit since the PR was opened. I'd recommend rebasing the whole PR onto the most recent 5.x branch.

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
```diff
@@ -156,9 +156,20 @@ private[sql] case class ElasticsearchRelation(parameters: Map[String, String], @
...
-
+    // Set fields to scroll over (_metadata is excluded, because it isn't a part of _source)
+    val scrollCSV = StringUtils.concatenate(filteredColumns.asInstanceOf[Array[Object]], StringUtils.DEFAULT_DELIMITER)
+    val requiredCSV = StringUtils.concatenate(requiredColumns.asInstanceOf[Array[Object]], StringUtils.DEFAULT_DELIMITER)
+    if (requiredCSV == cfg.getReadMetadataField)
+      // This is a small hack to avoid reading _source fields in case if only _metadata field is requested.
+      // By default, SearchRequestBuilder includes all fields, if INTERNAL_ES_TARGET_FIELDS is empty.
+      // To prevent us from querying useless data, we set _id field,
+      // which isn't a part of _source, but is allowed by SearchRequestBuilder
+      paramWithScan += (InternalConfigurationOptions.INTERNAL_ES_TARGET_FIELDS -> "_id")
```
jbaiera: I'm not sure I like relying on this hack. Perhaps we could find a way to signal that we don't want to read the `_source` field at all by sending `_source=false` with the request. That way we can avoid depending on obscure and potentially easy-to-break functionality in Elasticsearch.

jbaiera:
@drudim Are you still interested in moving forward with the PR?

drudim:
@jbaiera sure! I had a vacation, just got back. I am going to take look on it this weekend.

drudim:
>  I'd like to hear about what kind of testing problems you ran into as well to see if I can help out.

I tried to run:
```
./gradlew -Pscala=211 -Pdistro=hadoopYarn :elasticsearch-spark-20:integrationTest
```
and got:
```
:elasticsearch-spark-20:integrationTest

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkScalaSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkSQLScalaSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkSQLScalaSuite FAILED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkSQLSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkSQLSuite FAILED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkStreamingScalaSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkStreamingScalaSuite FAILED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkStreamingSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkStreamingSuite FAILED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.integration.SparkSuite STARTED

Gradle Test Run :elasticsearch-spark-20:integrationTest > Gradle Test Executor 2 > org.elasticsearch.spark.sql.UtilsTest STARTED

58 tests completed, 8 failed, 1 skipped
```

jbaiera:
When the build executes the test failures are logged to files in the build directory. You'll need to grab the error logs from there.

drudim:
Found the reason why `SparkSQLSuite.xml` was empty, it's derived from `AbstractJavaEsSparkSQLTest.xml`, so here is the error: https://gist.github.com/drudim/38faf2d3cc36b2e1a08f4779eee1ce8e

jbaiera:
I actually just got bit by that same exact problem this morning. The problem here is that we import Hadoop 1.2.1 for tests, which is a problem for Spark 2.0 in some cases. We have a shim implemented for the ShutdownHookManager in the `spark-20` `integTest` source root, but it does not include the `isShutdownInProgress` method on it. You may need to leave off the `-Pdistro=hadoopYarn` bit on the testing command for the time being. Alternatively you could try to see if it works by adding that method to our mock object.

drudim:
That's cool, will try to extend tests with my case later today. Let me know if you are ok with the new option to exclude `_source` field.

jbaiera:
@drudim Will do, I'll take a look at your changes and comments today.

drudim:
@jbaiera added the test for `_metadata` use-case. Ideally other parts should be covered as well (like sql-1.3 and `ScrollReader`), so I need more time.

drudim:
@jbaiera done:
- fixed the same bug in `sql-13`
- added integration test for `sql-13` and `sql-20`
- checked that `ScrollReader` is tested with empty source field: https://github.com/elastic/elasticsearch-hadoop/blob/master/mr/src/test/java/org/elasticsearch/hadoop/rest/ScrollReaderTest.java#L110
- haven't found a separate test for `SearchRequestBuilder` (to test `excludeSource`), but it's functionality indirectly tested by integration tests added in the scope of this PR.

spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
```diff
@@ -273,16 +273,21 @@ private[sql] object SchemaUtils {
...
+    val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
+    val sourceFields = SettingsUtils.determineSourceFields(settings)
+
+    // In case when user selected specific fields (we want to keep the same order in a spark sql row)
+    if (StringUtils.hasText(requiredFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
+
+    // In case when we read all fields including metadata
+    else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata)
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
```
jbaiera: This line is a little confusing to me: Is there ever a point in time where any of the `else` statements will be executed? I believe that with the current code `requiredFields` will ALWAYS be set with text, and if it is not, then `sourceFields` will also be empty. I think this could be simplified to just be:

```
val requiredFields = // get required fields
if (StringUtils.hasText(requiredFields)) {
  // set the rowInfo._1 properties
}
```

In the current state it makes me wonder if when we get to this code there is a chance that something prior in the connector wasn't set up entirely, but I don't see where it wouldn't be....

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
```diff
@@ -269,16 +268,21 @@ private[sql] object SchemaUtils {
...
+    if (StringUtils.hasText(requiredFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
+
+    // In case when we read all fields including metadata
+    else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata)
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields + StringUtils.DEFAULT_DELIMITER + settings.getReadMetadataField)
+
+    // In case when we read all fields without metadata
+    else if (StringUtils.hasText(sourceFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, sourceFields)
```
jbaiera: Curly braces please

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
```diff
@@ -269,16 +268,21 @@ private[sql] object SchemaUtils {
...
+
+    val requiredFields = settings.getProperty(Utils.DATA_SOURCE_REQUIRED_COLUMNS)
+    val sourceFields = SettingsUtils.determineSourceFields(settings)
+
+    // In case when user selected specific fields (we want to keep the same order in a spark sql row)
+    if (StringUtils.hasText(requiredFields))
+      rowInfo._1.setProperty(ROOT_LEVEL_NAME, requiredFields)
+
+    // In case when we read all fields including metadata
+    else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata)
```
jbaiera: Same confusion around simplifying the conditions here as the sql-13 package

spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/SchemaUtils.scala
drudim: Yes, you are right, one condition is enough. My original assumption was:
```
if (StringUtils.hasText(requiredFields)) {
   // for queries like dataFrame.select("specific", "fields").take(1)
} else if (StringUtils.hasText(sourceFields) && settings.getReadMetadata) {
   // for queries like dataFrame.take(1) when "es.read.metadata" is set to true
} else if (StringUtils.hasText(sourceFields) {
   // for queries like dataFrame.take(1) when "es.read.metadata" is set to false
}
```
But it isn't the case, because of the way how we set `requiredFields`.

drudim:
@jbaiera, ready for review:
- fixed style issues
- added assertions for `excludeSource` scenario (+tests)
- removed useless conditions in `SchemaUtils`

Issue #931, opened 2017-02-13
java.sql.Timestamp cannot be cast to scala.Tuple2


dmarkhas:
### What kind an issue is this?

- [x] Bug report. If you’ve found a bug, please provide a code snippet or test to reproduce it below.   
    The easier it is to track down the bug, the faster it is solved. 

### Issue description

The following exception is thrown when calling saveToEs on a DataFrame, if the operation is upsert, a script and params are provided and if one of the columns of the dataframe is a timestamp.

java.lang.ClassCastException: java.sql.Timestamp cannot be cast to scala.Tuple2

### Steps to reproduce

```
case class TestDoc(id: Int, SomeString: String, Date: java.sql.Timestamp)

val ts1 = new Timestamp(cal.getTime.getTime)
val ds = List(TestDoc(1234567,"SECOND",ts1)).toDF()
val script = "if (ctx._source.new_UpdateDate > new_UpdateDate) {ctx.op = 'none'} else {ctx._source.id = new_id;ctx._source.SomeString= new_SomeString;ctx._source.Date = new_Date}"
val params = "new_id:id,new_SomeString:SomeString,new_Date:Date"
ds.saveToEs("ipg4/doc", Map("es.mapping.id" -> "id", "es.write.operation" -> "upsert", "es.update.script"-> script, "es.update.script.params" -> params))
```

Strack trace:

```
java.lang.ClassCastException: java.sql.Timestamp cannot be cast to scala.Tuple2
	at org.elasticsearch.spark.sql.DataFrameValueWriter.write(DataFrameValueWriter.scala:55)
	at org.elasticsearch.hadoop.serialization.bulk.AbstractBulkFactory$FieldWriter.doWrite(AbstractBulkFactory.java:130)
	at org.elasticsearch.hadoop.serialization.bulk.AbstractBulkFactory$FieldWriter.write(AbstractBulkFactory.java:102)
	at org.elasticsearch.hadoop.serialization.bulk.TemplatedBulk.writeTemplate(TemplatedBulk.java:80)
	at org.elasticsearch.hadoop.serialization.bulk.TemplatedBulk.write(TemplatedBulk.java:56)
	at org.elasticsearch.hadoop.rest.RestRepository.writeToIndex(RestRepository.java:159)
	at org.elasticsearch.spark.rdd.EsRDDWriter.write(EsRDDWriter.scala:67)
[...]

PR #934, opened 2017-02-17
Support Date / Timestamp objects

dmarkhas:
Added support for handling Date and Timestamp objects - convert to epoch milliseconds

This addresses #931 

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/AbstractBulkFactory.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/AbstractBulkFactory.java
@@ -120,6 +120,12 @@ public abstract class AbstractBulkFactory implements BulkFactory {
 
                 pool.get().bytes(valueString);
             }
+            
+            else if (value instanceof Date) {
+                String valueString = (value == null ? "null": Long.toString(((Date) value).getTime()));
+                pool.get().bytes(valueString);
+            }
+            
             else if (value instanceof RawJson) {
                 pool.get().bytes(((RawJson) value).json());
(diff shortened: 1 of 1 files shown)


mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/AbstractBulkFactory.java
```diff
@@ -120,6 +121,12 @@ void doWrite(Object value) {
 
                 pool.get().bytes(valueString);
             }
+            
+            else if (value instanceof Date) {
+                String valueString = (value == null ? "null": Long.toString(((Date) value).getTime()));
+                pool.get().bytes(valueString);
+            }
```
jbaiera: This may fix the situation where a user tries to extract a date object for using in a parameter from the document, but I don't think it addresses the underlying problem that objects returned from the `DataFrameFieldExtractor` are not compatible with the `DataFrameValueWriter`.

dmarkhas: I agree that some refactoring is needed, but this small patch is what we applied in-house to be able to use the connector (we almost always use scripted updates, and almost always have datetime fields), so I'd appreciate if you can merge it as an interim fix before addressing the larger issue which may take longer.

jbaiera: I'll suspend this critique then. We're already doing this same logic for Strings, Numbers, Booleans, and nulls. I'll open a separate issue to address this

PR #937, opened 2017-02-20
when using Cascading local mode any attempt to load JSON data with es…

cwensel:
….input.json=true fails with a NoClassDefFoundError attempting to load non-existent Hadoop classes.

Cascading local mode was designed to be used with zero Hadoop MR, Tez, or Fink dependencies. 

if es.input.json=true, WritableBytesConverter will fail on a NoClassDefError attempting to load a Hadoop class.


jbaiera:
@cwensel Thanks for submitting the PR! This looks all good to me. I'll merge and forward port it to the 5.3 branch.

PR #940, opened 2017-02-23
Take Scala standard in spark support

takezoe:
This pull request includes following fix:

- Avoid procedure syntax method declaration
  - Because it brings potential of bug in the future code change. Procedure syntax is deprecated currently in the Scala standard and it will be removed in the future version of Scala.
- Avoid unnecessary `return` statement
  - `return` is translated into throwing an exception (and catching it) by Scala compiler, so we shouldn't use it in the performance reason. Also compiler can't inference a return type of the method if return is being used in the method.
- Avoid unnecessary `while` loop and mutable variable
  - Typically, we can write code safer by using immutable variables. `while` loop depends mutable flags in many cases, but it can be replaced with other way in Scala. It makes possible to eliminate mutable flags.

And I have questions about code in `ScalaValueReader.scala`, so I added inline comment to that code.

----

### DIFF
--- a/spark/core/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala
+++ b/spark/core/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala
@@ -59,6 +59,8 @@ import org.elasticsearch.hadoop.util.SettingsUtils
 import org.elasticsearch.hadoop.util.StringUtils
 import org.elasticsearch.hadoop.util.unit.Booleans
 
+import scala.annotation.tailrec
+
 class ScalaValueReader extends ValueReader with SettingsAware {
 
   var emptyAsNull: Boolean = false
@@ -71,11 +73,11 @@ class ScalaValueReader extends ValueReader with SettingsAware {
 
   def readValue(parser: Parser, value: String, esType: FieldType) = {
     if (esType == null) {
-      null
+      null // TODO Does this intend 'return null'?
     }
 
     if (parser.currentToken() == VALUE_NULL) {
-      nullValue()
+      nullValue() // TODO Does this intend 'return nullValue()'?
     }
 
     esType match {
@@ -186,18 +188,18 @@ class ScalaValueReader extends ValueReader with SettingsAware {
     new LinkedHashMap
   }
 
-  override def addToMap(map: AnyRef, key: AnyRef, value: Any) = {
+  override def addToMap(map: AnyRef, key: AnyRef, value: Any): Unit = {
     map.asInstanceOf[Map[AnyRef, Any]].put(key, value)
   }
 
-  override def wrapString(value: String) = {
+  override def wrapString(value: String): AnyRef = {
     value
   }
 
   def createArray(typ: FieldType): AnyRef = {
     nestedArrayLevel += 1
 
-    List.empty;
+    List.empty
   }
 
   override def addToArray(array: AnyRef, values: java.util.List[Object]): AnyRef = {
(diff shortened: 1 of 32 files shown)


spark/core/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala
```diff
@@ -71,11 +73,11 @@ class ScalaValueReader extends ValueReader with SettingsAware {
 
   def readValue(parser: Parser, value: String, esType: FieldType) = {
     if (esType == null) {
-      null
+      null // TODO Does this intend 'return null'?
```
takezoe: What is this null check intended? It looks no meaning...

spark/core/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala
```diff
@@ -71,11 +73,11 @@ class ScalaValueReader extends ValueReader with SettingsAware {
...
     }
 
     if (parser.currentToken() == VALUE_NULL) {
-      nullValue()
+      nullValue() // TODO Does this intend 'return nullValue()'?
```
takezoe: Same as the above.

jbaiera:
@takezoe thanks for the PR! I've been looking through the changes the last few days whenever I've had time. I'd like to do some digging regarding those confusing null value returns before merging, but it might take a little while with my schedule. Just wanted to pop in and give you a heads up about the status of the review. The changes look great so far!

takezoe:
@jbaiera Thanks for your comment. I'm trying to eliminate remained return statements, but it will take a while, so I will send them as a new pull pull request after this request is merged.

spark/core/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala
```diff
@@ -206,55 +208,41 @@ class ScalaValueReader extends ValueReader with SettingsAware {
...
+    _arrayDepth(potentialArray, 0)
   }
 
   def wrapArray(array: AnyRef, extraDepth: Int): AnyRef = {
       var arr = array
-      var i = 0
-      for (i <- 0 until extraDepth) {
-          arr = List(array)
+      for (_ <- 0 until extraDepth) {
+          arr = List(array) // TODO Should this line be arr = List(arr)?
```
takezoe: This line looks like a bug. Maybe correct code is `arr = List(arr)`?

takezoe:
@jbaiera I pushed more fix, and add a comment to the line that looks like a bug. Hope you will check them.

rabbitonweb:
Given the nulls might be a potential bug, maybe it is worth investigating?

takezoe:
Hi @jbaiera, what's the status of this pull request?

jbaiera:
Hi @takezoe, I'll be getting a proper review here very soon. My attention has been allocated toward other things at Elastic for the last month or so, but will be hopefully freeing up here in the next few days to return to here and clear out the backlog of things that need review. I seriously appreciate your patience on the matter!

takezoe:
@jbaiera Thanks, no problem if there are nothing blocking this pull request for now.

jbaiera:
@takezoe How did you end up tracking down all of these problems? I'm wondering if there's a way to establish a build check to make sure these sorts of problems don't arise during regular development going forward, like a code analyzer or setting the compiler to error on syntax warnings.

takezoe:
@jbaiera In fact, I found out them by my eyes. There are some lint tools for Scala but almost of them are provided as sbt plugin, so they can't be used in Gradle.

However Scala compiler has some options to print warnings, it's might be helpful to find parts of these problems. Perhaps it's possible to enable these options in Gradle as well. I will raise an another pull request to update Gradle configuration to specify Scala compiler options.

takezoe:
I created a new request to add Scala compiler options as #964.

takezoe:
@jbaiera Thanks for your review. All fixed and changed the pull request target to `5.x` branch.

spark/sql-20/src/main/scala/org/elasticsearch/spark/sql/RowValueReader.scala
```diff
@@ -52,7 +52,7 @@ private[sql] trait RowValueReader extends SettingsAware {
     }
   }
 
-  def addToBuffer(esRow: ScalaEsRow, key: AnyRef, value: Any) {
+  def addToBuffer(esRow: ScalaEsRow, key: AnyRef, value: Any): Unit = {
     val pos = esRow.rowOrder.indexOf(key.toString())
     if (pos < 0 || pos >= esRow.values.size) {
       // geo types allow fields which are ignored - need to skip these if they are not part of the schema
```
jbaiera: Thanks!

spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/RowValueReader.scala
```diff
@@ -52,15 +52,17 @@ private[sql] trait RowValueReader extends SettingsAware {
...
-  def addToBuffer(esRow: ScalaEsRow, key: AnyRef, value: Any) {
+  def addToBuffer(esRow: ScalaEsRow, key: AnyRef, value: Any): Unit = {
     val pos = esRow.rowOrder.indexOf(key.toString())
     if (pos < 0 || pos >= esRow.values.size) {
       // geo types allow fields which are ignored - need to skip these if they are not part of the schema
       if (pos < 0 && currentFieldIsGeo) {
-        return
+        ()
+      } else {
+        throw new EsHadoopIllegalStateException(s"Position for '$sparkRowField' not found in row; typically this is caused by a mapping inconsistency")
```
jbaiera: Thanks!

PR #954, opened 2017-04-05
Missing "JAR" keyword

ebuildy:
According https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Cli

It seems better to precise "ADD JAR"


jbaiera:
This LGTM. Not sure why the CI is reporting a failure. Let's see if poking it like this gets a clean re-test.

jbaiera:
Ah, I see what's going on here. It's trying to run the CI build against the 2.0 branch. Could you move these changes to the 5.x branch?

ebuildy:
It looks fixed at https://github.com/elastic/elasticsearch-hadoop/blob/5.x/docs/src/reference/asciidoc/core/hive.adoc. 

We can close this PR so?

jbaiera:
@ebuildy ah, you would be correct. It looks like this has been fixed since 2.3.0. I think we can safely close this then.

PR #994, opened 2017-05-22
options to ignore document missing or already exists exception

maoxiajun:
added two ignore options:
1. es.create.ignore.exists
2. es.update.ignore.document.missing

reason1 (es.create.ignore.exists) :
some scene like insert some docs but we don't want to overwrite already exists docs;

reason2 (es.update.ignore.document.missing):
some scene like we are updating some agg fields, and we don't care missing some already deleted docs

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -44,4 +44,4 @@ groovyVersion = 2.4.4
 # --------------------
 # Project wide version
 # --------------------
-version=5.0.0-beta1
+version=5.0.0.BUILD.SNAPSHOT


karmi:
Hi @maoxiajun, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-hadoop/pull/994.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

maoxiajun:
Hi Karel, I've not use 'git config' command to set my git user.name and user.email yet, sorry for that. I'll retry this after setting my git user.name and email. Thank you.

At 2017-05-22 15:47:58, "Karel Minarik" <notifications@github.com> wrote:


Hi @maoxiajun, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git commit. Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

maoxiajun:
Hi Karel, I've changed my Git commit e-mail and signed the CLA file.
I added two options to ignore two exceptions that's safe in some special scene:

1.
when reindex some index with statistical fields, we'll generate documents from hive, at which time, new documents are saved to elasticsearch by realtime computation, so we don't want these realtime documents be overwrited.

2.
still when update statistical fields, we don't care if some documents (like deleted goods) are missing, and we don't want update action be interrupted by document missing exception.

Thank you.

At 2017-05-22 15:47:58, "Karel Minarik" <notifications@github.com> wrote:


Hi @maoxiajun, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git commit. Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @maoxiajun, thanks!, the e-mail lookup works perfect now.

PR #996, opened 2017-05-22
add option to ignore document missing or already exists exception

maoxiajun:
add two options:
1. es.create.ignore.exists
2. es.update.ignore.document.missing

reasons:
1. we generate new docs in hive, but some realtime docs are already in elasticsearch, so we don't want to overwrite them when batch insert;
2. we generate statistical fields in hive, but realtime update action delete some invalid docs in elasticsearch, so just jump missing documents to finish update successfully.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -228,6 +228,13 @@ public interface ConfigurationOptions {
     String ES_OPERATION_DELETE = "delete";
     String ES_WRITE_OPERATION_DEFAULT = ES_OPERATION_INDEX;
 
+    // ignore document already exists exception, like we don't want to overwrite document that already exists when batch insert docs from hive
+    String ES_CREATE_IGNORE_EXISTS = "es.create.ignore.exists";
+    String ES_CREATE_IGNORE_EXISTS_DEFAULT = "false";
+    // ignore document missing exception, like we don't care deleted docs when update statistical fields
+    String ES_UPDATE_IGNORE_404 = "es.update.ignore.document.missing";
+    String ES_UPDATE_IGNORE_404_DEFAULT = "false";
+
     String ES_UPDATE_RETRY_ON_CONFLICT = "es.update.retry.on.conflict";
     String ES_UPDATE_RETRY_ON_CONFLICT_DEFAULT = "0";
 
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
@@ -279,6 +279,14 @@ public abstract class Settings {
         return Integer.parseInt(getProperty(ES_UPDATE_RETRY_ON_CONFLICT, ES_UPDATE_RETRY_ON_CONFLICT_DEFAULT));
     }
 
+    public boolean getUpdateIgnore404() {
+        return Boolean.parseBoolean(getProperty(ES_UPDATE_IGNORE_404, ES_UPDATE_IGNORE_404_DEFAULT));
+    }
+
+    public boolean getCreateIgnoreExists() {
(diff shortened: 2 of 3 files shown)


jbaiera:
1) Please avoid opening new feature PR's without an accompanying issue for discussing the ramifications.
2) What are the use cases that this solves that better support for the upsert operation wouldn't fix?

maoxiajun:
Hi, Sorry for that, it'll not happen again.
And for the "Upsert" option, like a e-store, I have a lot of goods for sale, for some reason I deleted some goods and I don't want it to be searched out and seen by others. But hive may still have the deleted goods in it, because we dump data from mysql once a day.
So if we use "Upsert " to update goods sold number with hive data, the deleted goods may be inserted into elasticsearch again, that's not what we want to see.




发自网易邮箱大师
On 05/25/2017 01:34, James Baiera wrote:
Please avoid opening new feature PR's without an accompanying issue for discussing the ramifications.
What are the use cases that this solves that better support for the upsert operation wouldn't fix?

jbaiera:
My main question is more along the lines of this: Is there existing functionality that we could be supporting in the connector to provide solutions to these use cases instead of promoting ignorance toward exceptions?

For instance, when doing an upsert in Elasticsearch, you can specify different document values for updating something versus creating something. Using this, you can leave the updating document empty to perform a no-op on the data if it already exists. In this way, you can ignore that a document already exists safely, without the hazard of ignoring an important error message.

Another example: You want to update data, but never insert it, and ignore that the document does not exist. You can perform a scripted upsert with an empty document body in the upsert section, and pass the full document you wish to update along in the script parameters. The script then chooses to ignore create operations and only performs updates, like so:

```
curl -H "Content-Type: application/json" -XPOST localhost:9500/test/data/5/_update?pretty -d '
{
  "scripted_upsert": true, 
  "script":{
    "inline":"if(ctx.op == \"create\") { ctx.op = \"none\" } else { ctx._source = params.doc }",
    "lang":"painless",
    "params": {
       "doc": {
          "field1" : "value1",
          "field2": "value2"
       }
    }
  }, 
  "upsert":{}
}'
```

Ultimately what I'm trying to say is that I'm uncomfortable adding code that ignores errors from Elasticsearch, and instead would rather add missing functionality to the connector that supports more complex use cases. In some cases that is not possible (see external versioning) but in cases of create and update operations, I would rather not simply ignore errors.

maoxiajun:
I think I was convinced by this, and I agree with you that we should not ignore exceptions silently.
But these options let me handle with safe exceptions (I thought them were ) conveniently, for some cluster in production, adding a inline script is not allowed.
So you can just close this PR if you think it is not necessary, I can use this in my customized release.




发自网易邮箱大师
On 05/26/2017 02:42, James Baiera wrote:

My main question is more along the lines of this: Is there existing functionality that we could be supporting in the connector to provide solutions to these use cases instead of promoting ignorance toward exceptions?

For instance, when doing an upsert in Elasticsearch, you can specify different document values for updating something versus creating something. Using this, you can leave the updating document empty to perform a no-op on the data if it already exists. In this way, you can ignore that a document already exists safely, without the hazard of ignoring an important error message.

Another example: You want to update data, but never insert it, and ignore that the document does not exist. You can perform a scripted upsert with an empty document body in the upsert section, and pass the full document you wish to update along in the script parameters. The script then chooses to ignore create operations and only performs updates, like so:

curl -H "Content-Type: application/json" -XPOST localhost:9500/test/data/5/_update?pretty -d '
{
  "scripted_upsert": true, 
  "script":{
    "inline":"if(ctx.op == \"create\") { ctx.op = \"none\" } else { ctx._source = params.doc }",
    "lang":"painless",
    "params": {
       "doc": {
          "field1" : "value1",
          "field2": "value2"
       }
    }
  }, 
  "upsert":{}
}'


Ultimately what I'm trying to say is that I'm uncomfortable adding code that ignores errors from Elasticsearch, and instead would rather add missing functionality to the connector that supports more complex use cases. In some cases that is not possible (see external versioning) but in cases of create and update operations, I would rather not simply ignore errors.

PR #1027, opened 2017-07-27
Remove YARN Integration

jbaiera:
Removal of the YARN Integration from Master and 6.0/6.x. Biggest items here are the documentation changes.


docs/src/reference/asciidoc/index.adoc
```diff
@@ -27,18 +27,26 @@
...
 
-repository-hdfs:: use HDFS as a back-end repository for doing snapshot/restore from/to {es}. For more information, refer to its https://www.elastic.co/guide/en/elasticsearch/plugins/master/repository-hdfs.html[home page]. For feature requests or bugs, please open an issue in the https://github.com/elastic/elasticsearch/issues[Elasticsearch repository] with the ":Plugin Repository HDFS" tag.
+repository-hdfs::
+Use HDFS as a back-end repository for doing snapshot/restore from/to {es}. For more information, refer to its
+https://www.elastic.co/guide/en/elasticsearch/plugins/master/repository-hdfs.html[home page]. For feature requests or
+bugs, please open an issue in the https://github.com/elastic/elasticsearch/issues[Elasticsearch repository] with the
+":Plugin Repository HDFS" tag.
 
-{ey}:: run {es} on top of YARN - see <<es-yarn>>. This project is in beta.
+{ey} [DISCONTINUED]::
```
acchen97: @dedemorton the es on yarn project had been deprecated and will be removed from the codebase in 6.0. discontinued sounded perfect to me, but thought you might know better. :)

acchen97: @dedemorton thanks for the input. The rationale behind having it persistent here is the fact that the project has been around for a couple years at this point, so we'd like to keep this discontinuation visible likely through the 6.x releases before removing it completely in 7.0.

Does that sound reasonable?

jbaiera:
I'm moving forward with merging this. If there's better terminology to use for the removal, we can apply it afterward.

docs/src/reference/asciidoc/index.adoc
dedemorton: @acchen97 I don't have a strong opinion either way since I don't own this doc. But I think your point could be made for a lot of features that we remove from the docs over time. Docs can get pretty bloated if we keep stuff like this around, and people forget to remove it. If we are  good about flagging deprecation early on and reminding users that the feature is going away, statements like this are unnecessary, IMO. While useful to users in the short term, these statements represent obsolete content that we need to maintain over time. IMO, it's better to train users to look in the release notes. I've been a tech writer for 20+ years so I've seen a lot of stuff like this die on the vine over the years. Anyhow, just something to keep in mind for the future.

PR #1033, opened 2017-08-07
Remove the last YARN references in documentation.

jbaiera:
To avoid confusion around the discontinuation of the YARN integration, we should just remove all references to it from the documentation. If users still require information about it, it will still be available in documentation of prior versions.


acchen97:
LGTM

PR #1035, opened 2017-08-11
Add -Xfatal-warnings option to scalac and fix all warnings

### LINKED ISSUE #974: Enable Fatal Warnings on Scala Compiler
Subtask of #969 

We should be catching and fixing common warnings in Scala that are reported by the compiler. Many of these are easy to miss and can evolve into bugs down the line if they aren't caught and fixed early. Enabling `-Xfatal-warnings` on the Scala compiler will fail builds if there are easily fixable Scala warnings, thus forcing developer(s) to handle them.

This depends on current warnings to be smashed out:
- [x] #970 Fix discarded return value Scala compiler warnings
- [x] #971 Fix unchecked type Scala compiler warnings.
- [x] #972 Fix dead code Scala compiler warnings
- [x] #973 Fix implicit conversion method Scala compiler warnings

takezoe:
Fixes #974

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -619,7 +619,8 @@ def baseSparkProject(target, sparkVersion) {
             "-Xfuture",
             "-Yno-adapted-args",
             "-Ywarn-dead-code",
-            "-Ywarn-numeric-widen"
+            "-Ywarn-numeric-widen",
+            "-Xfatal-warnings"
         ]
 
         sourceCompatibility = 1.6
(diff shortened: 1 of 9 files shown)


jbaiera:
@takezoe Sorry for the late review on this. I had some original comments pending from a while ago, but they never made it up. After examining the code again I don't see any issues, so I'm going to go ahead and approve this. Thanks so much (again) for your contributions toward the Spark side of the code base recently.

PR #1063, opened 2017-10-18
Add support for dropping bulk errors for particular HTTP status codes

aosagie:
Hi,
I run a 24/7 Spark Streaming to Elasticsearch application that processes around 4TB a day. Occasionally, a single bad document will get in and bring the whole thing down. Unfortunately, I have very little control over upstream parsing of the documents. So I'd like ES to just drop the individual document, log the error and continue.

I've seen several PRs in the past that dealt with ignoring various types of errors, but they never seem to get merged. Please let me know if there's anything I can do to clean up this PR and push this forward!

Thanks for a great product.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -245,6 +245,8 @@ public interface ConfigurationOptions {
     String ES_UPDATE_SCRIPT_PARAMS = "es.update.script.params";
     String ES_UPDATE_SCRIPT_PARAMS_JSON = "es.update.script.params.json";
 
+    String ES_BATCH_WRITE_DROP_ERROR_STATUSES = "es.write.drop.error.statuses";
+
     /** Output options **/
     String ES_OUTPUT_JSON = "es.output.json";
     String ES_OUTPUT_JSON_DEFAULT = "no";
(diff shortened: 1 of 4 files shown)


jbaiera:
Hi @aosagie, thanks for opening the PR.

As you mentioned, several PR's have been opened in the past that try to address this problem. The connector was originally built with batch processing in mind - deciding to fail early and often when it encounters issues with data rather than resisting the failures and continuing on. We have discussed this quite a bit internally, and are working toward designing a way for users to not only ignore errors that they know are mundane or irrelevant, but also to give users finer grained control over handling errors when it comes to serializing/deserializing data and Elasticsearch accepting/rejecting writes for various reasons.

I've opened an issue (#1064) to discuss the desired features for this.

cerebrotecnologico:
@aosagie I am in need of something similar. It seems this feature has been requested for a while. I found this thread:
https://discuss.elastic.co/t/catching-exceptions-from-savetoes-elasticsearch-spark/53745

I took the code from your fork. It provides what I need immediately. Nevertheless, I am thinking of a different approach,  I will attempt to make the ES connector to act as a filter, rejected records will be returned just as another RDD. Then I can direct those to the sink of my choosing (a file, a Kafka topic, /dev/null ).

jbaiera:
With the conclusion of error handlers for bulk failures (https://github.com/elastic/elasticsearch-hadoop/issues/1064), I believe that this is PR is no longer needed.

PR #1066, opened 2017-10-27
Connection leak that occurs when using the WAN support

shimamoto:
Hi,

I recently encountered a connection leak when using the elasticsearch-hadoop Spark Streaming support. After profiling the application, I found the connection leak is related to the WAN support. The details are listed below.

## Testing Application

This is a simple `saveToEs` application as follows.

```scala
object StreamingSample {

  def main(args: Array[String]): Unit = {
    val conf = new SparkConf().setAppName("sample streaming")
    val ssc = new StreamingContext(conf, Seconds(5))

    val rdd = ssc.sparkContext.makeRDD(Seq[Map[String, Int]]())
    new ConstantInputDStream(ssc, rdd).saveToEs {
      Map(
        ES_NODES    -> "localhost",
        ES_RESOURCE -> "demo/job",
        ES_NODES_WAN_ONLY -> "true"
      )
    }

    ssc.start
    ssc.awaitTermination
  }

}
```

## Environment

- Spark 2.2.0
- Elasticsearch 5.5.2
- OpenJDK Runtime Environment 1.8.0_131

## Failure symptoms

Under the top-level Spark directory, execute the following:

```sh
./bin/spark-submit --master local[2] --class StreamingSample <application-jar>
```

An error doesn't occur but connections are not released as follows.

```sh
root@7370be58c337:/# netstat -anlp|grep 9200|wc -l
5533
```

The above figures is the number of connections after about 2 hours. This figures will keep on increasing even after this.

When not using the WAN support, the number of connections is always constant.

## Identified root cause

The root cause is this:
https://github.com/elastic/elasticsearch-hadoop/blob/v5.5.2/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java#L614

The `randomNodeWrite` method creates a new instance of the `RestRepository` class. As a result, the previous created instance ([see](https://github.com/elastic/elasticsearch-hadoop/blob/v5.5.2/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java#L605)) is not being closed, which means that the connection is not released back into the pool.

## Solution

Fixes the code not to create a new instance of the `RestRepository` class when using the WAN support. Returns the already existed instance because the previous pinned node ([see](https://github.com/elastic/elasticsearch-hadoop/blob/v5.5.2/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java#L585)) is selected from the given nodes.

---

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/RestService.java
@@ -615,8 +615,14 @@ public abstract class RestService implements Serializable {
             }
         }
 
+        // if WAN mode is used, use an already selected node
         if (settings.getNodesWANOnly()) {
-            return randomNodeWrite(settings, currentInstance, resource, log);
+            String node = SettingsUtils.getPinnedNode(settings);
+            if (log.isDebugEnabled()) {
+                log.debug(String.format("Partition writer instance [%s] assigned to [%s]", currentInstance, node));
+            }
+
+            return repository;
         }
 
         // if client-nodes are used, simply use the underlying nodes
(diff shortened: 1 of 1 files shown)


jbaiera:
I will backport this change to the 6.X branches soon.

Issue #553, opened 2015-09-24
More support for `upsert` needed.


YiuTerran:
Currently `upsert` can't be configured to do some task just like in curl using HTTP API. 

For example, I want to `update` using doc A, but `create` if not exist with doc B. 
Or, i want `create` only if the doc not exist. And if it's already exist, do nothing(don't throw any exception).

I believe these requirements are important for some application.

costin:
More information in this [forum post](https://discuss.elastic.co/t/how-to-upsert-an-initial-value-into-elasticsearch-using-spark/29450).

rtrujill007:
:+1:

rtrujill007:
@costin @jbaiera Just to confirm if our workflow is covered by this issue:

 We want to have the ability to **Upsert** documents. If a document does not exist, then create it. If a document exists, then use an id field (preferably not the `_id` lets say another system maintained field called `objectId`) to perform the update. However we do not want to override or replace the auto generated `_id` field. 

Worse case scenario, we can also include the `_id` as part of the document (we would manually query ES and enhance the record with the `_id`) when we perform an **upsert**. It should be able to use the `_id` to perform the update as well. 

Thanks!

hanoch:
👍 +1

cclient:
+1
v: 6.0

now support

{
    "doc" : {
        "name" : "new_name"
    },
    "doc_as_upsert" : true
}
and 

{
    "script" : {
        "inline": "ctx._source.counter += params.count",
        "lang": "painless",
        "params" : {
            "count" : 4
        }
    },
    "upsert" : {
        "counter" : 1
    }
}

don't support while doc_as_upsert=false

like

{
"doc": {
    "lastupdatetime": "2015-09-16T18:00:00"
}
"upsert" : {
    "createtime": "2015-09-16T18:00:00"
    "lastupdatetime": "2015-09-16T18:00:00",
}
}

cclient:
a method do
{ "upsert": {}, "doc": {...} }

only minor modifications can do by self 

https://github.com/elastic/elasticsearch-hadoop/pull/1080/commits/27071fe4b842011a06126c15b8c0e22f75cdc42f

use demo

https://github.com/cclient/elasticsearch-spark-upsert-from-kafka

PR #1080, opened 2017-12-05
elasticsearch update doc

cclient:
update support 
`{
  "upsert": {},
  "doc": {...}
}`

demo
`case class ES_Upsert(kw_index: String, kw_type: String, id: String, date_idate: String, date_udate: String)`
`case class ES_Doc(date_udate: String)`
`case class ES_UpsertDoc(upsert: ES_Upsert, doc: ES_Doc)`

`  .saveToEs(Map[String, String](
      "es.resource" -> "{upsert.kw_index}/{upsert.kw_type}",
      "es.nodes" -> es,
      "es.input.json" -> "false",
      "es.update.doc" -> "true",
      "es.write.operation" -> "upsert",
      "es.mapping.exclude" -> "upsert.kw_index,upsert.kw_type,upsert.id",
      "es.mapping.id" -> "upsert.id"
    ))`

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -245,6 +245,9 @@ public interface ConfigurationOptions {
     String ES_UPDATE_SCRIPT_PARAMS = "es.update.script.params";
     String ES_UPDATE_SCRIPT_PARAMS_JSON = "es.update.script.params.json";
 
+    String ES_UPDATE_DOC = "es.update.doc";
+    String ES_UPDATE_DOC_DEFAULT = "no";
+    
     /** Output options **/
     String ES_OUTPUT_JSON = "es.output.json";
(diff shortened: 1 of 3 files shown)


jbaiera:
@cclient It seems that this PR is not completed? Will you be working on it further? One thing to keep in mind is that sending upserts versus updates is a feature all integrations use. This will need to work for frameworks other than Spark.

cclient:
@jbaiera Forgive my poor english,I only tested in spark,and won't working for other integrations, just wish it may be help for somebody who need '{ "upsert": {}, "doc": {...} }''

jbaiera:
@cclient We appreciate your willingness to help push the project forward. As described in #553, this issue affects all users for ES Hadoop. Unfortunately, we cannot accept a PR that only solves this problem for only a single integration. If you are willing to put work into a solution for the other integrations we can keep this PR open and move forward.

jbaiera:
Closing this out as it does not look like there will be further work applied to it. Happy to re-open if that changes.

Issue #1064, opened 2017-10-18
Introduce "Dead Letter Handlers" allowing interception of failures to handle/ignore them


jbaiera:
ES-Hadoop is a project that often embraces the fail-fast mantra, that exceptions and errors are often exceptional circumstances that need direct user interaction to fix and should not be ignored. This works great for batch jobs that are idempotent and easy to re-run, but the Hadoop ecosystem has since moved on to streaming applications being just as common as batch with patterns like lambda architecture and near-realtime processing.

As such, users are finding it difficult to use the connector as every time the connector experiences a non-recoverable exception such as in serializing malformed data, or data that is rejected by Elasticsearch under consistency guarantees.

We want to implement something close to Logstash's Dead Letter Queues, but persisting bad data can be a challenge when working on a distributed system like Hadoop. Logstash's Dead Letter Queues takes advantage of Logstash's internal binary formats and the fact that they are installed on a system with a (often) non-ephemeral filesystem. ES-Hadoop has no such internal binary format, opting to use each integration's data abstraction, and supports a plethora of integrations where HDFS or some other persistent storage is not always guaranteed to be available.

We are planning instead to implement something like a "Dead Letter _Handler_" instead of queue. Users would be able to implement interfaces for certain classes of errors or certain critical sections within the connector. These handlers would allow users to provide implementations that decide the connector's actions when potentially recoverable error states pertaining to data are encountered. Users would register the handlers with the connector by implementation name, and the connector will instantiate and configure the handlers at start up by means of reflection or SPI.

The project will also provide out of the box implementations of these handlers to cover the most common use cases, such as `Ignore errors and log them`, `Ignore all errors`, and the existing behavior `raise all exceptions`.

Users that wish to gain something similar to the Logstash Dead Letter Queue, would be able to implement their own handlers to persist the troubled data, such as sending it to HDFS, storing it in a queuing system like Kafka, or storing it on the local filesystem for example.

dmarkhas:
This is a good idea, but looking at all referenced pull requests they are all requesting a specific, rather simple, capability of ignoring specific HTTP response codes. Is it not possible to implement this feature first as a short-term solution until the proposed error handlers are implemented?

jbaiera:
@dmarkhas Thanks for your input here. We're apprehensive toward introducing a short term solution that would need to be very quickly deprecated and replaced when this solution arrives. 

[...]

PR #1095, opened 2018-01-18
Add "Dead Letter Handlers" for bulk write failures

jbaiera:
This PR adds APIs for users to provide custom code to handle errors encountered while the connector performs bulk write operations to Elasticsearch.

A common need that we have been seeing from users is that the connector lacks resiliency in the face of certain error states. Common cases are writing documents that fail due to having an outdated version number or garbage data sneaking by regular processing. In most cases, users want the ability to simply log the failed writes and continue processing. Our main apprehension to this change was that errors signal when something is wrong, and should be ignored with caution. Users should have the freedom to ignore errors that they expect to encounter, but should also have the tools to discern when an error is expected or not and behave accordingly. Some users may wish to collect and persist data that has experienced errors during writes. With this change, we are adding the tools for users to do all of this.

New user facing API's have been added in this PR. This includes the error handler interfaces, and new settings for specifying error handler class names and error handler properties. Error handlers are created at write initialization time and passed handler specific properties. Handlers are executed in the order that they were specified in the properties. A handler is given the opportunity to handle the error, mark the error as a valid error, or pass the error on to the next handler in the chain for processing. This creates a light weight pipeline for handling errors that allows users to compose handlers together to increase their job's resiliency to failure. When the job concludes, all handlers are closed so that they may clean up any resources they may have allocated.

Currently, the only place these error handlers are implemented is for bulk write failures, but the plan is to extend their use to handle other errors like serialization problems. The bulk flushing logic has been rewritten to account for handler code that may edit and retry data entries. Default error handler implementations for bulk write failures have been added to simulate the connector's current behavior. This includes two handlers: A handler that retries bulk entries a finite number of times that have been rejected due to load on Elasticsearch, and a handler that marks errors as valid errors, potentially halting the job. A third handler has been added that logs all bulk failures passed to it and marks the error has handled, thus ignoring the failure.


jbaiera:
Relates #1064

PR #1108, opened 2018-02-13
Add switchable findPartitions behavior.

valery-shinkevich:
For queries that process many indexes with many shards findSlicePartitions is the slowest.
There is a separate indexes for each day, every contains tens millions documents, i need to query six months for example.
findShardPartitions gets a lot more partitions, but it does not delay the execution.
Let, in my case, findShardPartitions will receive 18k partitions, but they will be processed in 10 minutes than I will wait until findSlicePartitions is performed in almost 2 hours.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -139,6 +139,9 @@ public interface ConfigurationOptions {
     String ES_INPUT_JSON = "es.input.json";
     String ES_INPUT_JSON_DEFAULT = "no";
 
+    String ES_INPUT_USE_SLICED_PARTITIONS = "es.input.use.sliced.partitions";
+    String ES_INPUT_USE_SLICED_PARTITIONS_DEFAULT = "true";
+
     /** Index settings */
     String ES_INDEX_AUTO_CREATE = "es.index.auto.create";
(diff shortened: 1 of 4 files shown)


mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
```diff
@@ -198,6 +198,9 @@
     String ES_SPARK_DATAFRAME_WRITE_NULL_VALUES = "es.spark.dataframe.write.null";
     String ES_SPARK_DATAFRAME_WRITE_NULL_VALUES_DEFAULT = "false";
 
+    String ES_SPARK_FIND_SLICE_PARTITIONS = "es.find.slice.partitions";
```
jbaiera: Lets move this up to the `/** Input options */` section and rename the property to something like "es.input.use.sliced.scroll" or ""es.input.use.sliced.partitions".

mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
```diff
@@ -579,6 +579,10 @@ public boolean getDataFrameWriteNullValues() {
         return Booleans.parseBoolean(getProperty(ES_SPARK_DATAFRAME_WRITE_NULL_VALUES, ES_SPARK_DATAFRAME_WRITE_NULL_VALUES_DEFAULT));
     }
 
+    public boolean getFindSlicePartitions() {
```
jbaiera: Reminder to match this method name up with the property name when it changes.

PR #1120, opened 2018-03-15
Overhaul the build system to be modular.

jbaiera:
Opening this PR while all the CI stuff is tested to not break immediately. Overhauling the build system so that it uses modular projects. Also included in this is automated scala variant compiling to use composite builds for compiling scala 2.10 artifacts.

Pull repository settings into build plugin.
Remove PMD plugin.
Remove Findbugs and Jacoco. We'll add it back later if we want it.
Add jbaiera to developer list.
Use https links in pom.
Point Hive docs to 1.2.2 because 1.2.1 is just gone.
Include propdeps in build source.
Burn S3 tasks to the ground.
Remove maven deployment for upload archives.
Ditch checksum operation on distzip.
Ditch dist.gradle script.
Add a root build plugin that defines and configures the master
archive tasks. Add an integration plugin that configures the archive
tasks on master with their own artifacts/sources/settings.
Remove test java from test resource dirs.
Convert spark13 to variant build. Revert back to spark 2.2.0.
Assemble should only build the archives. Remove the javadoc and source
jars from the archives configuration since we won't use upload archives
anymore. Pack task still packages all jars. Add distribution task. The
distribution task will take over the previous things that were heaped
on the assemble task, like the distZip, writing poms, and moving
artifacts.


koertkuipers:
it looks very clean now, but i am a little lost...

i used to change the urls to publish to in dist.grade, but that is gone, and i cannot find anything that looks like publishing urls anywhere else. where are they?

i used to run `./gradlew uploadArchives` but this task is gone? how do i publish the artifacts now?

thank you!

jbaiera:
The uploadArchives task is indeed gone now. Instead we rely on the CI systems to perform maven operations in the build since it's much easier on each project to have that centrally managed. Can you elaborate further on what you're trying to do? Maybe we can work out a way to get you what you need?

koertkuipers:
thanks!

i am trying to publish a version with some minor modifications to our inhouse artifact server (nexus, artifactory, etc.)

koertkuipers:
if the CI system performs these tasks you still need the CI system to call something like uploadArchives, no?

jbaiera:
For uploading to maven we have sets of credentials that CI is configured to use. As it stands, the ES-Hadoop project just generates the poms and their corresponding artifacts, and the CI system picks them up and does the uploading. I'm not well versed in how the central CI is set up so I can't really confidently comment much more on it.

The build does still use Gradle's built in Maven plugin for forming the pom. Perhaps it would be straightforward to modify the BuildPlugin.configureMaven() method to include a new upload archives task for your artifact server.

PR #1129, opened 2018-03-28
add additional error message

charsyam:
I met this error message, when I failed to create index.
 - I can see index with red status. 

so adding this message is more helpful to understand the situation and the error status.

- [O] I have signed the [Contributor License Agreement (CLA)][]


jbaiera:
Just to verify, was it the action of creating the index that cause the red status or was it already red and configured to ignore red status?

charsyam:
@jbaiera Actually, I gave bad option so, there are some timeouts.
after that, index was in wrong(red) status. and it shows the same error.

PR #1135, opened 2018-04-11
add sparkSession conf

hanbj:
val sparkConf = new SparkConf().setAppName("HiveToES").setMaster("local[*]")
        sparkConf.set("spark.sql.hive.metastorePartitionPruning", "false")
        sparkConf.set("es.index.auto.create", "true")
        val session1 = SparkSession.builder().config(sparkConf).getOrCreate()
        session1.sql("create table if not exists es_test using es OPTIONS (es.nodes 'localhost', es.port '9200',path 'hanbj')")
        val dataFrame = session1.sql("select * from es_test")
        dataFrame.show()

        val session2 = SparkSession.builder().config(sparkConf).getOrCreate()
        session2.conf.set("es.nodes", "localhost")
        session2.conf.set("es.port", "9201")
        val df = EsSparkSQL.esDF(session2, "ddposorderinfo")
        df.show()

In Spark2.x，SparkConf is a global configuration. Users can have multiple SparkSession and the configuration of each SparkSession can be different. This will be more flexible.

### DIFF
--- a/spark/sql-20/src/test/scala/org/elasticsearch/spark/sql/SparkSessionTest.scala
+++ b/spark/sql-20/src/test/scala/org/elasticsearch/spark/sql/SparkSessionTest.scala
@@ -5,7 +5,7 @@ import org.apache.spark.sql.SparkSession
 import org.junit.Assert._
 import org.junit.Test
 
-class EsSparkSQLTest {
+class SparkSessionTest {
 
   @Test
   def parameters(): Unit = {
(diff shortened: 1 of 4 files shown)


jbaiera:
@hanbj Thanks for the contribution - The CI build is failing right now due to issues with snapshots on master. While that comes back up, could you please add a test to make sure that the appropriate configurations are making their way into the connector settings?

hanbj:
@jbaiera I'm sorry, this is late test.

spark/sql-20/src/test/scala/org/elasticsearch/spark/sql/SparkSessionTest.scala
```diff
@@ -0,0 +1,26 @@
...
+
+class SparkSessionTest {
+
+  @Test
+  def parameters(): Unit = {
+    val sparkConf = new SparkConf().setAppName("session conf").setMaster("local[*]")
+    val session1 = SparkSession.builder().config(sparkConf).getOrCreate()
+    session1.conf.set("es.nodes", "127.0.0.1")
+    session1.conf.set("es.port", "9200")
+    assertTrue(session1.conf.get("es.nodes").equals("127.0.0.1"))
```
jbaiera: I'm not sure that this is testing any of the changes in this PR, maybe it makes sense to do this in the integration tests?

hanbj: The purpose of this test is to show that SparkSession can be configured differently to connect different clusters. This is the problem I have in my work. At work, we need to cache different instances of SparkSession, and use the instance to query different clusters and indexes when hitting cache.

In fact, I also modified the source code, adding two configurations of es.cluster.name and es.index.filter. es.cluster.name is to do the validation of the cluster name, and es.index.filter is to filter out the index I want to query in the large alias (the alias pointing to many indexes). In this way,each SparkSession instance (we cache) can query different indexes through the index of the es.index.filter configuration.

PR #1141, opened 2018-04-26
update the Slice bean which id and max should be accumulate in global

wankunde:
The number of slice should be ` (numAllDocs - 1) / maxDocsPerPartition + 1` ,not the number of the shard / maxDocsPerPartition.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -137,8 +137,13 @@ class BuildPlugin implements Plugin<Project>  {
             project.rootProject.ext.java8 = JavaVersion.current().isJava8Compatible()
 
             String javaHome = findJavaHome()
+            // Register the currently running JVM version under its version number.
+            final Map<Integer, String> javaVersions = [:]
+            javaVersions.put(Integer.parseInt(JavaVersion.current().getMajorVersion()), javaHome)
+
             project.rootProject.ext.javaHome = javaHome
             project.rootProject.ext.runtimeJavaHome = javaHome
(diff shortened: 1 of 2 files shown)


jimczi:
Can you explain why do you think this change is necessary ?
We compute the number of partitions per shard since each map is responsible for a slice inside a specific shard. So we need to ensure that the mappers responsible for a specific shard handle all partitions created for that shard (from 0 to max slice id). With your modification we'd miss documents since all slices wouldn't run on all shards.

wankunde:
Thank you very much for your reply.
I just made a test.
When `slice.max < slice.max`, only `slice.id = (shardId % slice.max)` has data.
When `slice.max > shards.size`, only `(slice.id % shards.size) =shardId`  has data.
I don't think that every shard needs to start so many map, many of which are unable to get data and end the task directly.

And I think `slice.max` should be caculate globally by `(numAllDocs - 1) / maxDocsPerPartition + 1` .

Am i right ?

jimczi:
Thanks for explaining @wankunde . The fact that some slices get data and some don't is a bug in es that we fixed in https://github.com/elastic/elasticsearch/pull/29533. Each slice query targets a single shard but the slicing is computed as if all shards were part of the query and that leads to empty bucket. The fix will be released in 6.4 and should resolve the issue with empty slices.

jimczi:
No we can't, this code has the same issue than this pr. Each shard must see all partitions created for that shard otherwise you'll miss documents. A working solution would be to decorrelate partitions and shards so that each partition targets all shards but that's a big change for 5x. Since this issue is fixed in 6.4 I'd advise to wait for this release and in the meantime you can reduce the cost of the empty slices by reusing the mapper jvm:
`mapred.job.reuse.jvm.num.tasks: 10`

wankunde:
OK, Thank you @jimczi  for your help!

Issue #1150, opened 2018-05-16
Consider chunking the docs to break up long pages and facilitate nav & retrieval


debadair:
Some of the pages are super long, and it's hard for users to navigate to the particular pieces of information they need. For example: https://www.elastic.co/guide/en/elasticsearch/hadoop/current/spark.html. 

Changing how the content is chunked or adding in-page nav would make it easier to use.

jbaiera:
#1151 is merged.

PR #1151, opened 2018-05-16
Breaking up the Spark documentation page into sub pages

jbaiera:
Have a page for each Spark integration.

There are a bunch of other things we could probably be doing to make our docs more readable and accessible.


jbaiera:
@debadair You mentioned an in-page nav in #1150—what is that and how would I apply something like that to the pages?

jbaiera:
I'm moving ahead with merging this in

PR #1154, opened 2018-05-28
An HTTP line is larger than 4096 bytes

hanbj:
val conf = new SparkConf().setAppName("EsTest").setMaster("local[*]")
        conf.set("spark.sql.shuffle.partitions", "6")
        conf.set("spark.default.parallelism", "6")
        conf.set("es.cluster.name", "common")
        conf.set("es.nodes", "localhost")
        conf.set("es.port", "9200")
        val sc = new SparkContext(conf)
        val sqlContext = new SQLContext(sc)
        val index = "le_plan_d/dpa"
        val query = "{\"query\":{\"match_all\":{}}}"
        val dataFrame = sqlContext.esDF(index, query)
        dataFrame.show(2)

My index has more than 600 fields. I execute a simple match_all query. 

The wrong information is as follows: 
Caused by: org.elasticsearch.hadoop.rest.EsHadoopInvalidRequest: An HTTP line is larger than 4096 bytes.
{"query":{"match_all":{}}}

So，Transfer to post mode.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
@@ -201,13 +201,6 @@ public class SearchRequestBuilder {
             uriParams.put("version", "");
         }
 
-        // override fields
-        if (StringUtils.hasText(fields)) {
-            uriParams.put("_source", HttpEncodingTools.concatenateAndUriEncode(StringUtils.tokenize(fields), StringUtils.DEFAULT_DELIMITER));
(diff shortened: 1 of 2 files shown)


jbaiera:
There seems to be some failing tests. Can you make sure that the tests complete successfully for you and comment when you have the PR updated?

hanbj:
This test is not very good. Because the HTTP restriction is within es, it is necessary to access es cluster validation.

wonderfuljamesjin:
Hi @hanbj and @jbaiera 

I am having the same issue,  can I know how to fix this issue.  thanks a lot

PR #1159, opened 2018-06-12
Support for custom MetadataExtractor within Spark flows

nerdynick:
Changes:
* Add support for a user defined MetadataExtractor within Spark. Also includes support for SettingsAware MetadataExtractor classes. 
* Expands the MetadataExtractor interface to recieve the Metadata Object during their work flow. 
* PerEntityPoolingMetadataExtractor expanded to support it's reuse within user defined MetadataExtractors.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -154,6 +154,8 @@ public interface ConfigurationOptions {
 
     /** Mapping types */
     String ES_MAPPING_DEFAULT_EXTRACTOR_CLASS = "es.mapping.default.extractor.class";
+    
+    String ES_MAPPING_METADATA_EXTRACTOR_CLASS = "es.mapping.metadata.extractor.class";
 
     String ES_MAPPING_ID = "es.mapping.id";
     String ES_MAPPING_ID_EXTRACTOR_CLASS = "es.mapping.id.extractor.class";
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
@@ -236,6 +236,10 @@ public abstract class Settings {
     public String getMappingDefaultClassExtractor() {
         return getProperty(ES_MAPPING_DEFAULT_EXTRACTOR_CLASS);
     }
+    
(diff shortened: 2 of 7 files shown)


jbaiera:
@nerdynick Is there an issue open for this? I'm not sure what the goal is here entirely. I can start reviewing it, but knowing what you're looking for here will help with the review process.

nerdynick:
@jbaiera No issue open. The basic goal was to allow users to also be able to provide MetadataExtractor's similar to how they can Index/Field Extractor's. The surrounding changes where in support for expanding the nature of MetadataExtractor to support user defined ones better. 

The use case I had was to be able to provide ESSpark with a Tuple2 where item 1 is an index pattern instead of the actual index and allowing that index pattern to be cached in a pool. The only way to gain that much level of control and information needed is by changing out the MetadataExtractor as the IndexExtractors do not have access to the Tuple2->item1 data. 

The only way I could see around this was hackie and would have involved mapping exclusions and overloading the value with more information to be thrown away. This just seemed simple and opened up more avenues for users to solve more complex use cases. 

Something I also noticed for another issue would be to move the Json transform layer into a MetdataExtractor instead of baking it in like it is currently. Seems like they are trying to achieve the same things but adding a whole bunch more layers to the mix to support the separation.

jbaiera:
I'm not entirely sold on this PR to be honest. After working on the new error handlers, its clear how much work it is to maintain these sorts of internal API's. We do not consider these interfaces to be a part of the public API for the connector, and thus they are not protected from breaking changes in between major releases as per semantic versioning rules. I feel like it would be incredibly trappy to expose them further.

nerdynick:
This PR is only exposing 1 more Extractor then is already exposed. As Field, Index, and Byte extractors are already exposed. As well as given the code that is/was already in place around MetadataExtractor it looked like the intent was already there with the Interface's existence and only 1 inheritance tree created (ScalaMetadataExtractor->PerEntityPoolingMetadataExtractor->MetadataExtractor).

I think an argument could be made for further exposing the use of PerEntityPoolingMetadataExtractor. But not to expose MetadataExtractor and its usage seems kinda limiting in the amount of gained support for a user to be able to optimize their pipelines flows and get the best performance out of it. With this PR I was able to better optimize and reduce each flow by 1-2sec per 10sec batch on SparkStreaming. As before I was having to do filter/split of data to 2 separate ES-Hadoop instance to achieve the same effect. This saved me from having to write my own Index Pattern logic. As well as still maintain isolation of all the logic around ES into one central ES layer instead of trying to weave it into my parse/transform layers.

Issue #1160, opened 2018-06-15
Reading from ES not including version metadata field


nathanielwendt:
### What kind an issue is this?

- [x] Bug report. If you’ve found a bug, please provide a code snippet or test to reproduce it below.   
    The easier it is to track down the bug, the faster it is solved. 

### Issue description

When reading data from ES, the _version field is not included despite setting configuration values for: `es.read.metadata` AND `es.read.metadata.version`.

### Steps to reproduce

Code:

```
    val queryString =
      """
        |{
        |  "query": {
        |    "term": { "catalog_id" : 2 }
        |  }
        |}
      """.stripMargin

    val config = Map(
      "es.read.metadata" -> "true",
      "es.read.metadata.version" -> "true",
      "es.read.field.as.array.include" -> "items.all_ids"
    )
    
    val es = spark.esDF("products/product", queryString, config)
    es.select("_metadata").show(1, false)
```

Output:

```
|Map(_score -> 3.9804614, _type -> product, _id -> p_1_0d0f1661a5b8b7f0170ca78c33ffc2823d62325fc0c8146e4a2a401f364bdc20, _index -> products_v2, sort -> Buffer(65051))|
``` 

### Version Info

OS:  OSX 10.11.6
JVM: Java 1.8
Hadoop/Spark: Spark 2.3.0
ES-Hadoop: 6.3.0
ES: 6.2.2

### Solution

I believe the issue is here: https://github.com/elastic/elasticsearch-hadoop/blob/3da13dd133042c077326cd0bc87b7597a1dab4af/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java#L203.  

With this line of code it builds a URI request like: `....&version&...`

There needs to be a value of "true" such that the URI request looks like `...&version=true&...`

I was able to correctly retrieve the _version field by setting breakpoints and correctly populating the uri params:

```
|Map(_score -> 3.9804614, _type -> product, _version -> 1, _id -> p_1_0d0f1661a5b8b7f0170ca78c33ffc2823d62325fc0c8146e4a2a401f364bdc20, _index -> products_v2, sort -> Buffer(65051))|
```

jbaiera:
Fixed with #1161

PR #1161, opened 2018-06-15
Bugfix - return _version field in metadata when requested

nathanielwendt:
Bugfix for: https://github.com/elastic/elasticsearch-hadoop/issues/1160

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/SearchRequestBuilder.java
@@ -200,7 +200,7 @@ public class SearchRequestBuilder {
         uriParams.put("scroll", String.valueOf(scroll.toString()));
         uriParams.put("size", String.valueOf(size));
         if (includeVersion) {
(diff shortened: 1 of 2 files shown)


nathanielwendt:
@jbaiera sorry for the delay on my end as well, I'm having trouble finding a good way to test this class as we only have public access to the `build` method that takes input of the large string created by the method I"ve updated in this PR.  Short of restructuring things quite a bit for this small change I"m not sure the best way to test it.  If you have an idea that I"m missing let me know and I"ll throw together a test class.

jbaiera:
@nathanielwendt Even something as simple as checking if the query string that is returned contains the substring `"version":"true"` would be fine in this case.

When accepting community PRs, I generally try to make time to do a second pass over the test code that is added since we can always be making things easier to test. I usually request a basic test case to be included in the event that I can't find time to make that second pass over the test code. In that case, at least there will be something basic there to catch issues.

nathanielwendt:
@jbaiera I've updated the code to include a quick test as you indicated.  Let me know if there is anything else I can do!

Issue #870, opened 2016-10-12
Unable to filter both text and keyword fields in same dataframe


jimmyjones2:
Elasticsearch & ES-Hadoop 5.0.0-beta1, Spark 2.0.0. Similar to #869 
```
curl -XPUT localhost:9200/test2 -d '{"mappings":{"test":{"properties":{"a":{"type":"text"},"b":{"type":"keyword"}}}}}'
curl -XPUT localhost:9200/test2/test/1 -d '{"a":"This is a sentance", "b":"Foo bar"}'

>>> df = sqlContext.read.format("org.elasticsearch.spark.sql").load("test2/test") 
>>> df.filter(df.a.like("This%")).show()
+------------------+-------+
|                 a|      b|
+------------------+-------+
|This is a sentance|Foo bar|
+------------------+-------+

>>> df.filter(df.b.like("Foo%")).show()
+---+---+
|  a|  b|
+---+---+
+---+---+

>>> df.filter((df.a.like("This%")) & df.b.like("Foo%")).show()
+---+---+
|  a|  b|
+---+---+
+---+---+

>>> df2 = sqlContext.read.format("org.elasticsearch.spark.sql").options(strict="true").load("test2/test")
>>> df2.filter(df2.a.like("This%")).show()
+---+---+
|  a|  b|
+---+---+
+---+---+

>>> df2.filter(df2.b.like("Foo%")).show()
+------------------+-------+
|                 a|      b|
+------------------+-------+
|This is a sentance|Foo bar|
+------------------+-------+

>>> df2.filter((df2.a.like("This%")) & df2.b.like("Foo%")).show()
+---+---+
|  a|  b|
+---+---+
+---+---+
```
Is it possible to the strict parameter to be automatically applied depending on the mapping/type of the field? Without this, it is impossible to filter on two fields if one is text and one is keyword.

jbaiera:
As with #869 we don't currently take the Elasticsearch mapping into account for these strict pushdown overrides, only the Spark Schema. Since going from Elasticsearch schema to Spark Schema can be lossy (for instance keyword+text both map to StringType), both should be considered for whether or not something should be a strict pushdown.

I'm wondering if `strict` could be extended to explicitly mention fields that should be made strict, or perhaps adding another option, like `strict.fields.include` or `strict.fields.exclude`, and just have `strict` set all of them to auto include?

jimmyjones2:
[...]

PR #1168, opened 2018-06-28
change match to terms

hanbj:
I created an external table for hive and I intend to read data from ES through HQL. 
When I run the following code. I can't find the data.

val sparkConf = new SparkConf()
val builder = SparkSession.builder().config(sparkConf).enableHiveSupport()
val spark1 = builder.getOrCreateBigquerySparkSession()
spark1.sql("set spark.bigquery.fullScan.enable=true")
spark1.sql("use olap")
spark1.sql("show tables").collect.foreach(println)
spark1.sparkContext.setLogLevel("fatal")  
spark1.sql("select * from table_name where city_code in ('010', '791') limit 100").show

But the following two HQL can all find data.
1. select * from table_name where city_code in ('010')
2. select * from table_name where city_code in ('791')

Reading source found that:  
When HQL was parsed into DSL,  in ('010','791') was parsed into "match": {"city_code": "010 791"}，

The right thing to do is： 
    "terms": {
      "city_code": ["010","791"]
    }

### DIFF
--- a/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
+++ b/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
@@ -455,10 +455,10 @@ private[sql] case class ElasticsearchRelation(parameters: Map[String, String], @
        StringUtils.EMPTY
      } else  {
        if (SettingsUtils.isEs50(cfg)) {
-         s"""{"match":{"$attribute":${strings.mkString("\"", " ", "\"")}}}"""
+         s"""{"terms":{"$attribute":${strings.mkString("[\"", "\",\"", "\"]")}}}"""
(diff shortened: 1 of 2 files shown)


jbaiera:
@hanbj thank you for opening this PR, but I don't think this is the way we should approach this issue. 

The SQL translation layer assumes that all queries written are match queries by default. If you want to perform terms queries, you should set [`strict` to `true` on your dataframe settings](https://www.elastic.co/guide/en/elasticsearch/hadoop/current/spark.html#spark-data-sources). That said, using `strict` mode has its drawbacks in that you can't mix an match `term` and `match` queries: https://github.com/elastic/elasticsearch-hadoop/issues/870

hanbj:
@jbaiera Sorry, I look at the logic again.  You're right.

PR #1169, opened 2018-06-28
Respect user provided schema

viirya:
When there is user provided schema, we should respect it.

### DIFF
--- a/spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
+++ b/spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
@@ -373,6 +373,33 @@ class AbstractScalaEsScalaSparkSQL(prefix: String, readMetadata: jl.Boolean, pus
   }
 
   @Test
+  def testArrayValueWithUserSchema {
+    val index = wrapIndex("sparksql-test-array-value-with-schema")
+    val indexAndType = s"$index/data"
+    RestUtils.touch(index)
+
+    // add some data
+    val jsonDoc = """{"array" : [1, 2, 4, 5] }"""
+    sc.makeRDD(Seq(jsonDoc)).saveJsonToEs(indexAndType)
+    RestUtils.refresh(index)
+
+    val newCfg = collection.mutable.Map(cfg.toSeq: _*) += (ES_READ_FIELD_AS_ARRAY_INCLUDE -> "array")
+    val schemaInt = StructType(StructField("array", ArrayType(IntegerType)) :: Nil)
+
+    val df = sqc.read.options(newCfg).format("org.elasticsearch.spark.sql")
+      .schema(schemaInt).load(indexAndType)
+
+    assertEquals("array", df.schema("array").dataType.typeName)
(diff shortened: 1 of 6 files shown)


viirya:
@jbaiera Thanks! I'm not familiar with the spark integration tests here, but I will try to add it.

spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
```diff
@@ -372,6 +372,33 @@ class AbstractScalaEsScalaSparkSQL(prefix: String, readMetadata: jl.Boolean, pus
...
+    // add some data
+    val jsonDoc = """{"array" : [1, 2, 4, 5] }"""
+    sc.makeRDD(Seq(jsonDoc)).saveJsonToEs(indexAndType)
+    RestUtils.refresh(index)
+
+    val newCfg = collection.mutable.Map(cfg.toSeq: _*) += (ES_READ_FIELD_AS_ARRAY_INCLUDE -> "array")
+    val schemaInt = StructType(StructField("array", ArrayType(IntegerType)) :: Nil)
+
+    val df = sqc.read.options(newCfg).format("org.elasticsearch.spark.sql")
+      .schema(schemaInt).load(indexAndType)
```
jbaiera: After checking this out and running it I found that the test does not actually pass. It looks like it passed in CI because of an issue in the build process.

I made some changes to the test locally, namely adding a field to the jsonDoc and omitting it from the schema. The connector fails because it cannot determine the spark field type for the data when it reads it. This makes me think that there are some deeper issues that need to be addressed before we can accept this kind of PR, namely around how we parse the data returned from Elasticsearch.

viirya:
> Do you have any specific use cases here that you are looking to satisfy with this change?

@jbaiera thanks for looking into this. Yes, I have used Spark to extract data from Elasticsearch cluster, however the schema of one field is not correct. Even I don't need the data of this field and only select fields I need, `elasticsearch-hadoop` still can't allow me to pull the data and throw exception when parsing the schema.

I'm not sure if this incorrect schema can be corrected or not by changing Elasticsearch's settings. Even it is feasible, I still have no permission to change the settings. Thus, I need to specify schema manually to pull the data.

jbaiera:
> Yes, I have used Spark to extract data from Elasticsearch cluster, however the schema of one field is not correct.

@viirya Is this due to the field not being stored the way you expect in Elasticsearch? In the test I saw that you provide a schema stating that the field should be an int, but by default Elasticsearch treats regular numbers as longs. Would you be able to make changes or conversions to the data in Spark to meet your expected data types? If not, what kinds of fields are you seeing this happen to?

> Even I don't need the data of this field and only select fields I need, elasticsearch-hadoop still can't allow me to pull the data and throw exception when parsing the schema.

Do you have a reproduction of those issues? I'm mostly confused since when you perform a select statement and discard a column in Spark SQL, it should exclude that column from the list of fields to read from Elasticsearch.

jbaiera:
@viirya Just checking in here to see if you're still interested in continuing work on this PR.

viirya:
@jbaiera Thanks for pinging me. It's a while and I left previous company. Let me check this and see if it can be continued.

jbaiera:
Closing this since there's been no movement on here. Happy to reopen if need be.

PR #1185, opened 2018-08-08
make sure the RestRepository is initialized when calling delete() method

jaratec:
This pull request is a fix for the following bug. When trying to create an index from a Spark job, by using the overwrite flag, the method delete() of RestRepository is called. However the bulk processor/writer are not initialized, thus causing an exception.

PS: I have signed the [Contributor License Agreement (CLA)][]

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/RestRepository.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/RestRepository.java
@@ -369,6 +369,8 @@ public class RestRepository implements Closeable, StatsAware {
             client.delete(resources.getResourceWrite().index() + "/" + resources.getResourceWrite().type());
         }
         else {
+            lazyInitWriting();
+
             // try first a blind delete by query (since the plugin might be installed)
             try {
                 client.delete(resources.getResourceWrite().index() + "/" + resources.getResourceWrite().type() + "/_query?q=*");


jbaiera:
Hi @jaratec, thanks for opening a PR for this. Could you include an integration test in the `AbstractScalaEsScalaSparkSQL` files (both 2.0 and 1.6 sub projects) that exercises the change?

jaratec:
Hello,
the bug is in the class RestRepository, of mr subproject. It seems to me that an integration test in this subproject would be more closer to the change. However, the integration suite for the rest package has an ignored annotation (org.elasticsearch.hadoop.integration.rest.RestSuite). When removing the annotation and running the tests, they are failing. It seems that in order to test RestRepository, a running ElasticSearch instance is required. I haven't figured out how to do that. Maybe you can point me to a test that does that. A test requiring a running ElasticSearch instance would be an integration test (in integ part, not in test part), right?
I have looked at the scala subprojects, at AbstractScalaEsScalaSparkSQL, however this one is not the same package as DefaultSource and ElasticsearchRelation that would excercise the code in RestRepository. These two classes have private package visibility, and are not visible in AbstractScalaEsScalaSparkSQL.

jbaiera:
@jaratec If you add a test to `AbstractScalaEsScalaSparkSQL` that reproduces the error then the code change should be tested enough to be accepted.

The RestSuite does indeed require a locally running Elasticsearch node. You can stand one up with Gradle using:
```
./gradlew elasticsearch-hadoop-mr:esCluster#wait -x elasticsearch-hadoop-mr:esCluster#stop
```
and tear it down with 
```
./gradlew elasticsearch-hadoop-mr:esCluster#stop
```

jbaiera:
No response on this PR for a while. The bug also looks like it's fixed in master now. Closing.

PR #1215, opened 2018-11-05
Add a build qualifier system property to the build

jbaiera:
Allows build processes to pick up a qualifier to add into the version
at build time instead of depending on it from a version file. By
default, it will continue to use alpha1 as the qualifier until the ES
build changes the master branch to ship an unqualified snapshot version
during the 7 prerelease period.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,16 +10,31 @@ boolean localRepo = project.getProperties().containsKey("localRepo")
 
 Properties props = new Properties()
 props.load(project.file('esh-version.properties').newDataInputStream())
-version = props.getProperty('eshadoop')
+String eshVersion = props.getProperty('eshadoop')
 String esVersion = props.getProperty('elasticsearch')
+
+// TODO: This default should be removed when ES stops using alpha1 as its default qualifier
+String qualifier = System.getProperty("build.version_qualifier", "alpha1")
+if (qualifier.isEmpty() == false) {
+    if (qualifier.matches("(alpha|beta|rc)\\d+") == false) {
+        throw new IllegalStateException("Invalid qualifier: " + qualifier)
+    }
+    eshVersion += "-" + qualifier
+    esVersion += "-" + qualifier
+}
 
 // determine if we're building a snapshot or not (by default we will be)
 boolean snapshot = "true".equals(System.getProperty("build.snapshot", "true"))
 if (snapshot) {
     // we update the version property to reflect if we are building a snapshot or a release build
-    version += "-SNAPSHOT"
-    props.put("eshadoop", version)
+    eshVersion += "-SNAPSHOT"
+    esVersion += "-SNAPSHOT"
 }
+
+props.put("eshadoop", eshVersion)
+props.put("elasticsearch", eshVersion)
+
+// determine if we're building a prerelease or candidate (alpha1/rc1/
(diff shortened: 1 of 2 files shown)


buildSrc/build.gradle
```diff
@@ -45,7 +60,7 @@ dependencies {
     compile 'org.springframework.build.gradle:propdeps-plugin:0.0.7'
 
     if (localRepo) {
-        compile name: "build-tools-${version}"
+        compile name: "build-tools-${esVersion}"
     } else {
         compile group: 'org.elasticsearch.gradle', name: 'build-tools', version: esVersion
```
jakelandis: This forces the qualified ES to exist in the remote repo prior to allowing this build to succeed... is that intentional ?
For example
```
./gradlew clean assemble -Dbuild.version_qualifier=alpha1 -Dbuild.snapshot=false
```
will not build until ES alpha1 (not snapshot) is available on the remote repo, in the prior version, it was controlled solely by the property which, (to me) is unclear if this is a change in the behavior/workflow for the release.

buildSrc/build.gradle
```diff
@@ -10,17 +10,32 @@ boolean localRepo = project.getProperties().containsKey("localRepo")
...
-    version += "-SNAPSHOT"
-    props.put("eshadoop", version)
+    eshVersion += "-SNAPSHOT"
+    esVersion += "-SNAPSHOT"
 }
 
+props.put("eshadoop", eshVersion)
+props.put("elasticsearch", eshVersion)
+
+// determine if we're building a prerelease or candidate (alpha1/rc1/
```
jakelandis: orphaned comment ?

buildSrc/build.gradle
```diff
@@ -10,17 +10,32 @@ boolean localRepo = project.getProperties().containsKey("localRepo")
 
 Properties props = new Properties()
 props.load(project.file('esh-version.properties').newDataInputStream())
-version = props.getProperty('eshadoop')
+String eshVersion = props.getProperty('eshadoop')
 String esVersion = props.getProperty('elasticsearch')
 
+// TODO: This default should be removed when ES stops using alpha1 as its default qualifier
+String qualifier = System.getProperty("build.version_qualifier", "alpha1")
```
jakelandis: Does the release manager need to be aware of this system property ?

buildSrc/build.gradle
jbaiera: Further up in the script the `localRepo` variable is set from the system properties. The localRepo property is used to configure a local directory as a repository for resolving project dependencies. During the release process, Elasticsearch is built first and the resulting build tools artifact is copied to the localRepo directory in the ES-Hadoop project. This way ES-Hadoop can depend on the exact same artifact that will be released by the same build. Those setup steps and scripts are managed outside this script by the release automation.

buildSrc/build.gradle
jbaiera: There's a separate PR open for updating the release automation with the new qualifier code.

buildSrc/build.gradle
jbaiera: Yup! Pushed 8e20684

buildSrc/build.gradle
jakelandis: Thanks for the explains!

PR #1216, opened 2018-11-16
Get canonical file path in Version's classpath normalization

jbaiera:
ES-Hadoop attempts to verify that there is only one ES-Hadoop jar on the classpath when the connector first comes online. When running Spark on a local Yarn installation on Mac, the node manager that launches the job container will incorrectly add the `/tmp` AND the `/private/tmp` paths to the classpath. For Mac, `/tmp` is a symlink to `/private/tmp` which leads to the ES-Hadoop jar registering twice on the classpath, thus tripping the single jar check.

This change simplifies the path normalization to canonicalize each jar's file path, thus resolving any symbolic or hard filesystem links. Each URL considered for validation is bucketed by its canonical file path, and any bucket that contains at least one jar file is counted once toward the total count of ES-Hadoop jars on the classpath.

### DIFF
--- a/mr/src/itest/java/org/elasticsearch/hadoop/VersionTest.java
+++ b/mr/src/itest/java/org/elasticsearch/hadoop/VersionTest.java
@@ -113,7 +113,7 @@ public class VersionTest {
         File privateTmpOtherDir = new File(privateTmpDir, "other");
         mkdirSafe(privateTmpOtherDir);
         File tmpDir = new File(linkTest, "tmp");
-        linkDir(privateTmpDir, tmpDir);
+        linkDir(tmpDir, privateTmpDir);
 
         // File in original directory
         File testFile = new File(privateTmpDir, "test");
@@ -166,7 +166,7 @@ public class VersionTest {
         }
     }
 
-    private void linkDir(File privateTmpDir, File tmpDir) throws IOException {
+    private void linkDir(File tmpDir, File privateTmpDir) throws IOException {
         try {
             Files.createSymbolicLink(tmpDir.toPath(), privateTmpDir.toPath());
         } catch (UnsupportedOperationException e) {
--- a/mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
@@ -216,6 +216,11 @@ public abstract class IOUtils {
 
     /**
      * Convert either a file or jar url into a local canonical file, or null if the file is a different scheme.
+     * @param fileURL the url to resolve to a canonical file.
+     * @return null if given URL is null, not using the jar scheme, or not using the file scheme. Otherwise, returns the
+     * String path to the local canonical file.
+     * @throws URISyntaxException If the given URL cannot be transformed into a URI
+     * @throws IOException If the jar cannot be read or if the canonical file cannot be determined
      */
     public static String toCanonicalFilePath(URL fileURL) throws URISyntaxException, IOException {
         if (fileURL == null) {
@@ -223,12 +228,12 @@ public abstract class IOUtils {
         }
 
         // Only handle jar: and file: schemes
-        if (!fileURL.getProtocol().equals("jar") && !fileURL.getProtocol().equals("file")) {
+        if (!"jar".equals(fileURL.getProtocol()) && !"file".equals(fileURL.getProtocol())) {
             return null;
         }
 
         // Parse the jar file location from the jar url. Doesn't open any resources.
-        if (fileURL.getProtocol().equals("jar")) {
+        if ("jar".equals(fileURL.getProtocol())) {
             JarURLConnection jarURLConnection = (JarURLConnection) fileURL.openConnection();
             fileURL = jarURLConnection.getJarFileURL();
         }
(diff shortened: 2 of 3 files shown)


mr/src/itest/java/org/elasticsearch/hadoop/VersionTest.java
```diff
@@ -42,12 +56,121 @@ public void testNormalize() throws Exception {
...
+        }
+    }
+
+    private void mkdirSafe(File dir) throws IOException {
+        if (!dir.mkdirs()) {
+            throw new IOException("Could not create " + dir + " folder");
+        }
+    }
+
+    private void linkDir(File privateTmpDir, File tmpDir) throws IOException {
```
jakelandis: nit: try to avoid switching the order of arguments (e.g. tmpDir is second arg on this method, but first arg on the one it is wrapping)

mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
```diff
@@ -209,4 +213,33 @@ private static BytesArray unwrapStreamBuffer(InputStream in) {
...
+    /**
+     * Convert either a file or jar url into a local canonical file, or null if the file is a different scheme.
+     */
+    public static String toCanonicalFilePath(URL fileURL) throws URISyntaxException, IOException {
+        if (fileURL == null) {
+            return null;
+        }
+
+        // Only handle jar: and file: schemes
+        if (!fileURL.getProtocol().equals("jar") && !fileURL.getProtocol().equals("file")) {
```
jakelandis: nit: prefer "constant".equals over .equals("constant") to avoid NPE .. which is not an issue here just a good practice.

mr/src/itest/java/org/elasticsearch/hadoop/VersionTest.java
```diff
@@ -42,12 +56,121 @@ public void testNormalize() throws Exception {
...
+            String norm = IOUtils.toCanonicalFilePath(url);
             normalized.add(norm);
         }
 
-        System.out.println(normalized);
         assertEquals(1, normalized.size());
     }
+
+    @Test
+    public void testWindowsNormalizeCanonicalPath() throws Exception {
```
jakelandis: thank you for all the comments it makes these tests very readable.

mr/src/main/java/org/elasticsearch/hadoop/util/Version.java
```diff
@@ -52,19 +54,45 @@
...
+                    List<URL> pathURLs = normalized.get(canonicalPath);
+                    if (pathURLs == null) {
+                        pathURLs = new ArrayList<URL>();
+                        normalized.put(canonicalPath, pathURLs);
+                    }
+                    pathURLs.add(url);
+                } catch (URISyntaxException e) {
+                    String message = "Could not parse classpath resource URI: " + url.toString();
+                    LogFactory.getLog(Version.class).fatal(message, e);
+                    throw new Error(message, e);
```
jakelandis: Why throw `Error` instead of `Exception` or `RuntimeException` ? Error is generally discouraged unless you also own the catching the Error and then only for framework level code as a last resort to prevent crashing the JVM.

Also, any specific reason to log AND throw, usually the thing catching has the responsibility to log as to avoid duplicated logs.

mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
```diff
@@ -209,4 +213,33 @@ private static BytesArray unwrapStreamBuffer(InputStream in) {
         }
         return null;
     }
+
+    /**
+     * Convert either a file or jar url into a local canonical file, or null if the file is a different scheme.
+     */
+    public static String toCanonicalFilePath(URL fileURL) throws URISyntaxException, IOException {
```
jakelandis: should we have proper java for public methods ? (e.g. \@params , \@returns \@throws )

jbaiera: This should be addressed in eb6652a but github isn't marking it as outdated

jbaiera:
```
spark-submit --conf spark.es.port=9200  --conf spark.es.resource=spark/_doc --conf spark.es.nodes.wan.only=true --class "co.elastic.sample.spark.WordCount" --master yarn --jars /tmp/elasticsearch-hadoop-6.5.0.jar target/hadoop-sample-1.0-SNAPSHOT.jar sampledata/wordcount/one.txt
```
Circling back here, I was able to reproduce on Spark 2.4.0 (homebrew) and 2.3.1 (tarball) running on both `client` and `cluster` deploy modes on YARN 2.7.7 (tarball).

My reproduction:
```
HADOOP_CONF_DIR=hadoop/hadoop-2.7.7/etc/hadoop/ spark-submit --class org.elasticsearch.hadoop.qa.kerberos.spark.LoadToES --master yarn --deploy-mode cluster --jars elasticsearch-hadoop-7.0.0-alpha1-SNAPSHOT-testing.jar --conf spark.es.resource=qa_kerberos_spark_data/_doc --conf spark.es.nodes=localhost:9200 --conf spark.load.field.names=number,name,url,picture,@timestamp,tag kerberos-7.0.0-alpha1-SNAPSHOT.jar /data/artists/artists.dat
```

PR #1218, opened 2018-11-20
bump 6.5 branch version to 6.5.2

droberts195:


jbaiera:
CI fails because there is no snapshot build yet. I'm moving forward with merging it.

PR #1221, opened 2018-12-07
Upgrade gradle5 jdk11

atorok:
build-tools now targets Java 10 ( class version 54 ). 
Building with older versions of java no longer works,
while the build doesn't seem to apply the plugin it does uses classes form build tools like 
`VersionProperties` which can't be loaded when the build is ran with older JDKs. 

This PR makes it possible to build using java 11 while setting target version to 8 so it will still work on older JVMs (this is the oldest ES supports).  
I'm not sure if this is the right  target, We seem to support Spark `2.0+` which from a quick search seems to support java 7 up  to `2.2`.


jbaiera:
Integration tests are failing at the moment with these changes on the local build-tools snapshot. It's tripping up on some low level Hive + JDBC + JDO mess related to the newer java runtime. I know that there are some Hadoop incompatibilities as well from code that parses java versions in static blocks. We will have to modify the build to use different java runtimes for the tests like ES does.

jbaiera:
I think this should just be closed out at this point. Getting the build on to jdk 11 is going to require the removal of some older supported framework versions as they simply do not support running on top of java versions above jdk 8.

Now that https://github.com/elastic/elasticsearch/pull/36453 is merged, the build should pass once more. If we feel that this should still be discussed, then I'm fine with re-opening. Until then I'll close this out.

PR #1245, opened 2019-02-06
Disable testing in kerberos project when localRepo is set

jbaiera:
When building release candidates, ES-Hadoop receives its build tools dependency via the release automation so that it has the exact version of the artifact that will be released. This means that any testing code that relies on Elasticsearch versions that are yet to be released needs to be skipped in the release candidate generation as it is both non-production testing code and stuck in a chicken-egg dependency conundrum.

With the addition of the qa kerberos project, this problem becomes a little more difficult, as elasticsearch cluster fixture plugins are not applied when running with a local repository of dependencies. Additionally, the qa-kerberos project depends on integration test code for building artifacts that as previously mentioned is not able to compile due to issues with resolving non existing artifact versions.

This change checks if a local repo is configured and skips the process of configuring the qa kerberos project, thus avoiding the issues with a missing Elasticsearch cluster fixture, as well as transfers some integration test specific classes to the qa kerberos project to break its dependency on the test runtimes in other projects for the purposes of compilation.

### DIFF
--- a/qa/kerberos/build.gradle
+++ b/qa/kerberos/build.gradle
@@ -40,17 +40,21 @@ configurations {
     stormDeps
 }
 
+boolean localRepo = project.getProperties().containsKey("localRepo")
+
 dependencies {
     compile project(":elasticsearch-hadoop-mr")
-    compile project(":elasticsearch-hadoop-mr").sourceSets.itest.runtimeClasspath
-
     compile project(":elasticsearch-hadoop-cascading")
+    compile project(":elasticsearch-storm")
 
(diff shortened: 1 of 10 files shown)


qa/kerberos/build.gradle
```diff
@@ -78,643 +82,366 @@ dependencies {
 // Testing jars
 // =============================================================================
 
-// Build uber cascading jar for testing
```
jbaiera: Apologies in advance for the bad diff rendering - this is just the rest of the file wrapped in an if statement

PR #1246, opened 2019-02-07
Update kerberos file test to account for multiple output files

jbaiera:
Updates the `AbstractClusterVerificationTests` to be sensitive to the fact that there may be multiple files that contain output from the test jobs in the qa kerberos project.

### DIFF
--- a/qa/kerberos/src/itest/java/org/elasticsearch/hadoop/qa/kerberos/AbstractClusterVerificationTests.java
+++ b/qa/kerberos/src/itest/java/org/elasticsearch/hadoop/qa/kerberos/AbstractClusterVerificationTests.java
@@ -39,21 +39,21 @@ public class AbstractClusterVerificationTests {
     @Parameters
     public static Collection<Object[]> params() {
         List<Object[]> params = new ArrayList<>();
-        params.add(new Object[]{"mr",        "part-m-00000", 345, true});
-        params.add(new Object[]{"spark",     "part-00000",   345, true});
+        params.add(new Object[]{"mr",        "part-m-", 345, true});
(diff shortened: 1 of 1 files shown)


qa/kerberos/src/itest/java/org/elasticsearch/hadoop/qa/kerberos/AbstractClusterVerificationTests.java
```diff
@@ -69,14 +69,16 @@ public void testOutputContents() throws Exception {
                     successMarker.exists());
         }
 
-        File dataFile = new File(integrationDataDir, dataFileName);
-        Assert.assertTrue("Expected data file [" + dataFile + "] to exist but could not find it.", dataFile.exists());
-        BufferedReader reader = new BufferedReader(new InputStreamReader(new FileInputStream(dataFile)));
         int lines = 0;
-        while (reader.readLine() != null) {
-            lines++;
+        for (File directoryFile : integrationDataDir.listFiles()) {
```
dakrone: Might be easier to do something like

```java
final LongAdder lines = new LongAdder();
Files.walk(Paths.get("foo"))
    .filter(p -> p.startsWith(prefix))
    .forEach(p -> {
        try (Stream<String> line = Files.lines(p)) {
            line.forEach(l -> lines.increment());
        } catch (IOException e) {
            fail("got: " + e);
        }
    });
```

(totally up to you though)

jbaiera: I left it simplified without lambdas to make the backport to the 6.7 branch easier, it's still on java 6. ☹️

PR #1257, opened 2019-02-27
Fix the integration test suites usage of typeless api calls

jbaiera:
In the 7.0 line there are changes to ensure that 7.0 respects the deprecation of the usage of Elasticsearch types in requests. This change ensures that the existing test code that depends on typeless API's continues to function in a manner that is backwards compatible with older versions for testing purposes.

The plan at the moment is to remove the types completely in the 8.0 line with a later PR. The backwards compatibility sensitive changes are in this PR for the benefit of running 7.x tests against 6.x clusters.

### DIFF
--- a/hive/src/itest/java/org/elasticsearch/hadoop/integration/hive/AbstractHiveSaveTest.java
+++ b/hive/src/itest/java/org/elasticsearch/hadoop/integration/hive/AbstractHiveSaveTest.java
@@ -695,10 +695,7 @@ public class AbstractHiveSaveTest {
         System.out.println(server.execute(selectTest));
         System.out.println(server.execute(insert));
 
-        String docEndpoint = resource("hive-fieldexclude", "data");
-        if (TestUtils.isTypelessVersion(targetVersion)) {
-            docEndpoint = docEndpoint + "/_doc";
-        }
+        String docEndpoint = docEndpoint("hive-fieldexclude", "data");
         String string = RestUtils.get(docEndpoint + "/1");
         assertThat(string, containsString("MALICE"));
 
@@ -711,6 +708,14 @@ public class AbstractHiveSaveTest {
     private String resource(String index, String type) {
         if (TestUtils.isTypelessVersion(targetVersion)) {
             return index;
+        } else {
+            return index + "/" + type;
+        }
+    }
+
+    private String docEndpoint(String index, String type) {
+        if (TestUtils.isTypelessVersion(targetVersion)) {
+            return index + "/_doc";
         } else {
             return index + "/" + type;
         }
--- a/hive/src/itest/java/org/elasticsearch/hadoop/integration/hive/AbstractHiveSaveTest.java
+++ b/hive/src/itest/java/org/elasticsearch/hadoop/integration/hive/AbstractHiveSaveTest.java
@@ -696,7 +696,7 @@ public class AbstractHiveSaveTest {
         System.out.println(server.execute(insert));
 
         String docEndpoint = resource("hive-fieldexclude", "data");
-        if (targetVersion.onOrAfter(EsMajorVersion.V_8_X)) {
+        if (TestUtils.isTypelessVersion(targetVersion)) {
             docEndpoint = docEndpoint + "/_doc";
         }
         String string = RestUtils.get(docEndpoint + "/1");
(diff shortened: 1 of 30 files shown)


hive/src/itest/java/org/elasticsearch/hadoop/integration/hive/AbstractHiveSaveTest.java
```diff
@@ -696,15 +695,27 @@ public void testMappingExclude() throws Exception {
         System.out.println(server.execute(selectTest));
         System.out.println(server.execute(insert));
 
-        String string = RestUtils.get("hive-fieldexclude/data/1");
+        String docEndpoint = resource("hive-fieldexclude", "data");
+        if (TestUtils.isTypelessVersion(targetVersion)) {
+            docEndpoint = docEndpoint + "/_doc";
+        }
```
jakelandis: Is this right ?

`hive-fieldexclude/data/_doc` ?  or should it be conditionally only `hive-fieldexclude/_doc` ?

mr/src/test/java/org/elasticsearch/hadoop/util/TestUtils.java
```diff
@@ -50,7 +50,17 @@ public static boolean delete(File file) {
...
+        try {
+            return client.mainInfo();
+        } finally {
+            client.close();
+        }
+    }
+
+    public static boolean isTypelessVersion(EsMajorVersion version) {
+        // Types have been deprecated in 7.0.0, and will be removed at a later date
+        return version.onOrAfter(EsMajorVersion.V_7_X);
```
jakelandis: minor: can you move the `resource` and `docEndpoint` to this method to dry up the methods ?

pig/src/itest/java/org/elasticsearch/hadoop/integration/pig/AbstractPigExtraTests.java
```diff
@@ -120,10 +124,27 @@ public void testTupleSaving() throws Exception {
                 "answers = LOAD 'src/itest/resources/tuple.txt' using PigStorage(',') as (id:int, parentId:int, score:int);" +
                 "grouped = GROUP answers by id;" +
                 "ILLUSTRATE grouped;" +
-                "STORE grouped into 'pig-test-tuple-structure/data' using org.elasticsearch.hadoop.pig.EsStorage('es.mapping.pig.tuple.use.field.names = true');";
+                "STORE grouped into '"+resource("pig-test-tuple-structure", "data")+"' using org.elasticsearch.hadoop.pig.EsStorage('es.mapping.pig.tuple.use.field.names = true');";
```
jakelandis: nit: String.format may allow this read better ? (opinion)

jakelandis:
This raises an interesting point w/r/t types and data written by ES.  Do you forsee issues for users upgrading to 7.0.0. For example, for Hive since since the table property is only set on the create table (which can happen in 6.7), will the integration still work in 7.0.0 (I assume yes, but with depreciations). That may be bad example, but are there integrations (pig?) that may be more sensitive to removal of sending types by default. Is there a plan for 8.0 once types are no longer allowed ?

jbaiera:
Thanks for the feedback @jakelandis 

I'm under the impression that the 7.X line will continue to support type names in API's (with deprecations and explicit options required) until 8.X lands. With that in mind, I view ES-Hadoop the same way as we would for any client code that would need to be updated to be typeless:

* 6.X: 
Types are optional for read operations, and are still required for write operations.
* 7.X:
Types are still accepted for write operations, but deprecations will be logged when types are present. It is expected that users remove the types from their table definitions and job configurations during this period of time.
* 8.X:
Types will be removed and will no longer supported. ES-Hadoop throws exceptions if type names are present in the resource while it is operating against ES clusters of 8.0 and up.

pig/src/itest/java/org/elasticsearch/hadoop/integration/pig/AbstractPigExtraTests.java
jbaiera: Should we just do that for all the Pig scripts? There are portions in other classes where multiple lines are injected using string concat, and replacing it with "%s" in those spots doesn't feel much cleaner...

Java needs string interpolation.

jakelandis: nah, just wanted to ask the question if you thought it would read better. it's good as-is

PR #1264, opened 2019-03-15
Add third party dependency license and notice files

jbaiera:
This PR removes transitive dependency resolution from all compile, provided, or , instead depending on explicit inclusion of dependencies in build files. Each dependency's LICENSE and NOTICE file from their project source control has been added to the licenses directories of each project. Tasks have been added that ensure that each dependency has an accompanying license file available on the project. Additionally, files that have been repackaged or derived from third parties have been marked with official notices and their original license headers.


jakelandis:
@jbaiera - Can you point out any areas that we should pay particular attention to, or ways to help validate this ?  Also, can we assume that the licenses and header texts are copy/pasted ?

jbaiera:
@jakelandis The license and notice file contents are all copied from the dependency projects as close as I could get to the release tag/branch connected to the version we use. The sha files are all generated. Most important stuff is probably the BuildPlugin and build file changes, as well as the changes from 43ff11a

I think a closer look might be helpful for any of the Jackson related dependencies since we depend on the codehaus version and not the fasterxml version. The codehaus mirror I found for 1.8.8 did not seem to have a license marked on it, but Maven seems to assert that it is Apache 2.0 licensed, so I treated it as such.

tomcallahan:
@jbaiera can you show an example of what the NOTICE.txt would look like in the resulting product distribution? I.e., the one that would ultimately appear in the es-hadoop zip file

jbaiera:
@tomcallahan [Here](https://github.com/elastic/elasticsearch-hadoop/files/3432429/NOTICE.txt) is the resulting NOTICE.txt file that is generated when the project is built using the current revision.

PR #1270, opened 2019-04-03
Docs: Drop inline callouts

### LINKED ISSUE #1262: Drop inline callouts from docs
We're working to switch the core of the docs build tools from the no-longer-maintained AsciiDoc to the actively-maintained Asciidoctor. For the most part Asciidoctor is compatible with AsciiDoc or we've made it compatible with extensions. But in one case we've decided not to make it compatible because it is both difficult and doesn't make the docs a lot better. It *does* require that we change a bunch of asciidoc files, sadly. In particular, we need to drop "inline callouts". An example from elasticsearch-hadoop:
```
.Map/Reduce
["source","xml",subs="attributes"]
----
<dependency>
  <groupId>org.elasticsearch</groupId>
  <artifactId>elasticsearch-hadoop-mr<1></artifactId>
  <version>{ver}</version>
</dependency>
----

<1> 'mr' artifact
```

It can be made to be supported with:
```
.Map/Reduce
["source","xml",subs="attributes"]
----
<dependency>
  <groupId>org.elasticsearch</groupId>
  <artifactId>elasticsearch-hadoop-mr</artifactId> <1>
  <version>{ver}</version>
</dependency>
----

<1> 'mr' artifact
```

nik9000:
Drops the inline callouts from the docs. This is when you write `<1>`
anywhere but the end of a line. Asciidoctor doesn't support them and
we'd very much like to move to Asciidoctor to generate the docs because
it is being actively maintained.

I've taken a shot at removing all of the callouts and everything
compiles but I suspect it doesn't look quite right.

Closes #1262

### DIFF
--- a/docs/src/reference/asciidoc/core/hive.adoc
+++ b/docs/src/reference/asciidoc/core/hive.adoc
@@ -59,7 +59,7 @@ When using Hive, one can use `TBLPROPERTIES` to specify the <<configuration,conf
 CREATE EXTERNAL TABLE artists (...)
 STORED BY 'org.elasticsearch.hadoop.hive.EsStorageHandler'
 TBLPROPERTIES('es.resource' = 'radio/artists',
-              'es.index.auto.create' = 'false') <1>;
+              'es.index.auto.create' = 'false'); <1>
 ----
 
 <1> {eh} setting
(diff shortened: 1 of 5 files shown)


nik9000:
@jbaiera, I converted all of the inline callouts into non-inline callouts but I don't know that I preserved the accuracy of the examples. Could you have a look?

You can now build the docs with ` ../docs/build_docs --doc docs/src/reference/asciidoc/index.adoc --asciidoctor --out ../built_docs --open`. They build about 3x faster with `--asciidoctor` and that is what we're trying to move towards. It will fail to build if it hits any inline callouts.

nik9000:
@jbaiera these are the branches that need to have this change:
```
[ master, 7.x, { 7.0: 7.0.0-rc2 }, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0 ]
```

Which is a whole bunch. Sorry about that.....

PR #1275, opened 2019-04-08
Docs: Clean up for asciidoctor

nik9000:
This makes three changes in preparation for switching the docs to
Asciidoctor:
1. Fixes a broken link. As a side effect this fixes a missing emphasis
in Asciidoctor that was caused by parsing issues with the `_` in the old
link.
2. Fixes an `added` macro that renders "funny" in Asciidoctor.
3. Replace a tab in a code example with spaces. AsciiDoc was doing this
automatically but Asciidoctor preserves the tab. We don't need the tab.


nik9000:
Merged and backported!

Issue #1276, opened 2019-04-13
include_type_name=true is added to ES 5.x queries eventhough it doesn't support that


adriancole:
It seems the code that adds `include_type_name=true` should guard on versions of ES that support it. Right now, it dies on ES 5.x due to this. I'm not sure which version after 6.3.2 broke this.

```
org.elasticsearch.hadoop.rest.EsHadoopInvalidRequest: org.elasticsearch.hadoop.rest.EsHadoopRemoteException: illegal_argument_exception: request [/test_zipkin:span-2019-04-13/_mapping/span] contains unrecognized parameter: [include_type_name]
null

	at org.elasticsearch.hadoop.rest.RestClient.checkResponse(RestClient.java:465)
	at org.elasticsearch.hadoop.rest.RestClient.execute(RestClient.java:422)
	at org.elasticsearch.hadoop.rest.RestClient.execute(RestClient.java:384)
	at org.elasticsearch.hadoop.rest.RestClient.execute(RestClient.java:388)
	at org.elasticsearch.hadoop.rest.RestClient.get(RestClient.java:168)
	at org.elasticsearch.hadoop.rest.RestClient.getMappings(RestClient.java:315)
	at org.elasticsearch.hadoop.rest.RestClient.getMappings(RestClient.java:305)
	at org.elasticsearch.hadoop.rest.RestRepository.getMappings(RestRepository.java:293)
	at org.elasticsearch.hadoop.rest.RestService.findPartitions(RestService.java:252)
```

```bash
$ curl http://localhost:32801/
{
  "name" : "lY1n3bH",
  "cluster_name" : "elasticsearch",
  "cluster_uuid" : "PDUcXXSeQf-RrFawFZPqzQ",
  "version" : {
    "number" : "5.6.16",
    "build_hash" : "3a740d1",
    "build_date" : "2019-03-13T15:33:36.565Z",
    "build_snapshot" : false,
    "lucene_version" : "6.6.1"
  },
  "tagline" : "You Know, for Search"
}

$ curl http://localhost:32801/test_zipkin-2019-04-13/_mapping/span?include_type_name=true
[...]

PR #1284, opened 2019-05-06
Fix the mapping client call on ES 5.x and earlier.

jbaiera:
This PR removes the include_type_names parameter from mapping requests made to versions of ES that do not recognize it. Included are a set of test updates that ensures that test runs against ES 5.x complete correctly.

Relates #1276


adriancole:
I noticed this is being bumped around instead of getting into a release. can you please release it?

jbaiera:
@adriancole This change just barely missed the last release window. I can confirm that it will be present in the listed release versions.

adriancole:
ok thanks.. will watch out for it then!

Issue #1277, opened 2019-04-13
ES 2.x broken since 6.4.0


adriancole:
I believe df2e67893911abf1c61bc7ff3cc1beac99e71704 broke elasticsearch 2.x support. It introduces the following state check which is invalid in 2.x when you initially encounter a result:

```java
        if (responseHits > 0) {
            return new Scroll(scrollId, totalHits, results, responseHits, skippedHits);
        } else {
            // Scroll had no hits in the response, it must have concluded.
            return new Scroll(scrollId, totalHits, true);
        }
```

The "Scroll had no hits in the response, it must have concluded." assumption is invalid. As you'll notice, the initial response from ES 2.x has no contents in the hits section. Flipping the `true` to `false` works for ES 2.x, but I think it would be better to pass in the version and perhaps 5.x forward use the assumption that it is indeed complete.

```
4:12:46,847 TRACE [Executor task launch worker for task 4] commonshttp.CommonsHttpTransport - Tx [POST]@[localhost:32898][test_zipkin:span-2019-04-13/span/_search]?[search_type=scan&scroll=5m&size=50&preference=_shards%3A3%3B_local] w/ payload [{"query":{"match_all":{}}}]
14:12:46,850 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "POST /test_zipkin:span-2019-04-13/span/_search?search_type=scan&scroll=5m&size=50&preference=_shards%3A3%3B_local HTTP/1.1[\r][\n]"
14:12:46,855 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "Content-Type: application/json[\r][\n]"
14:12:46,857 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "Accept: application/json[\r][\n]"
14:12:46,857 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "User-Agent: Jakarta Commons-HttpClient/3.1[\r][\n]"
14:12:46,857 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "Host: localhost:32898[\r][\n]"
14:12:46,858 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "Content-Length: 26[\r][\n]"
14:12:46,858 DEBUG [Executor task launch worker for task 4] httpclient.Wire - >> "[\r][\n]"
[...]

PR #1287, opened 2019-05-09
Fix incorrect scroll termination for ES 2.x

jbaiera:
When initiating a scroll in 5.x and above, an empty initial response indicates that the scroll/slice being read has no data in it. When ES-Hadoop reads this initial empty response it assumes that the scroll is empty and concludes the read operation. In ES 2.x however, the initial scroll response is always empty, and the resulting behavior is that ES-Hadoop reads nothing from the index at all. This PR adds in a switch when reading from ES 2.x and below that if the initial scroll request is empty or marked as concluded, it is modified to allow for subsequent requests to determine if it is actually empty or not.


jbaiera:
Relates #1277

jbaiera:
Manual tests pass for ES versions 8.0.0-SNAPSHOT, 7.0.1, 6.5.2, 5.6.2, 2.4.6

PR #1309, opened 2019-06-25
version bump to 7.2.1

jakelandis:
(ci is expected to fail)

### DIFF
--- a/docs/src/reference/asciidoc/index.adoc
+++ b/docs/src/reference/asciidoc/index.adoc
@@ -9,7 +9,7 @@
 :krb:   Kerberos
 :ey:	Elasticsearch on YARN
 :description: Reference documentation of {eh}
-:ver:	7.2.1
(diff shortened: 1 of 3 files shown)


docs/src/reference/asciidoc/index.adoc
```diff
@@ -9,9 +9,9 @@
 :krb:   Kerberos
 :ey:	Elasticsearch on YARN
 :description: Reference documentation of {eh}
-:ver:	7.2.0
-:ver-d: 7.2.0-SNAPSHOT
-:es-v:	7.2.0-SNAPSHOT
+:ver:	7.2.1
```
jbaiera: I think this line needs to remain 7.2.0 since that's the currently released version. This is a weird detail that only applies on the first release of a minor version.

```suggestion
:ver:	7.2.0
```

PR #1320, opened 2019-07-17
Fix Lucene's snapshot version

matriv:
Lucene's snapshot version was wrong which resulted in failed builds.

1. ES version was used instead of Lucene version in the BuildPlugin`
2. Lucene's snapshot version was wrong in the properties file


jbaiera:
LGTM, thanks for opening this!

Issue #1321, opened 2019-07-17
Find Lucene version automatically


matriv:
Replace the manual setting of Lucene's version in `esh-version.properties` file with a mechanism that finds it automatically through ES version.

Currently when Lucene is updated the version should be updated in this repo as well manually, which can be the source of build failures.

See for example #1320.

atorok:
The hadoop build already uses `build-tools` so could be looking at version.properties ( there's also a `VersionProperties` class ) to figure out the lucene version.

matriv:
@atorok When I tried the VersionProperties I get `8.0.0` for lucene:
![Screenshot 2019-07-17 at 16 52 21](https://user-images.githubusercontent.com/5058131/61381139-4878cf00-a8b3-11e9-84ae-5b251924cdce.png)

atorok:
Maybe the wrong version of build-tools is being pulled in ? Or a snapshot that doesn't have the update yet.

matriv:
I tried `7.2.0.` version of build tools but still I get `8.0.0` from `VersionProperties.lucene`.
Do you mean could something be wrong with es version?

jbaiera:
The manual versioning for Lucene is tied to the fact that the build-tools artifact is no longer java 8 compliant. The decision at the time of it breaking was to hard code ES-Hadoop to use an older version of build-tools in order to unblock daily snapshots. The escape valves for lucene and build tools versions were added at that time. I've gone through and tried undoing the original hack but it seems that the build tools are still java 11 only on 7.x.

jbaiera:
Closing this as I don't think we pull the lucene artifacts for the project any longer.

PR #1323, opened 2019-07-19
Use new snapshot of build-tools with Java 8 support

matriv:
Part of 7.x build-tools now have Java 8 support, so VersionProperties
can be used to retrieve Lucene's snapshot version and therefore there is
no need to set it manually to the properties file.

Fixes: #1321

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -58,6 +58,7 @@ dependencies {
 
     // Provided/Optional Dependencies
     compile 'org.springframework.build.gradle:propdeps-plugin:0.0.7'
+    implementation 'org.apache.logging.log4j:log4j-core:2.11.1'
 
     // TODO: Un-comment this once we can safely depend on build-tools again.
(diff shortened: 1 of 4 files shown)


buildSrc/build.gradle
```diff
@@ -58,6 +58,7 @@ dependencies {
 
     // Provided/Optional Dependencies
     compile 'org.springframework.build.gradle:propdeps-plugin:0.0.7'
+    implementation 'org.apache.logging.log4j:log4j-core:2.11.1'
```
matriv: Hope this doesn't affect something.
Had to add it because of the following error:
```
> Task :buildSrc:compileGroovy
error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor not found
startup failed:
Compilation failed; see the compiler error output for details.
```

PR #1324, opened 2019-07-22
Implement delete operation

vincentarnaud90:
Implement delete operation and tests. This is a duplicate from https://github.com/elastic/elasticsearch-hadoop/pull/1242 with testing.

- [X] I have signed the [Contributor License Agreement (CLA)][]


vincentarnaud90:
Hello, is the pull request still missing something ?

jbaiera:
@vincentarnaud90 Thanks for re-commenting on this PR, as it slipped through my todo list. I'll get a review up for it soon.

mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/DeleteBulkFactory.java
```diff
@@ -0,0 +1,99 @@
...
+import org.elasticsearch.hadoop.util.StringUtils;
+
+import java.util.List;
+
+public class DeleteBulkFactory extends AbstractBulkFactory {
+
+    public static final class NoDataWriter implements ValueWriter<Object> {
+
+        @Override
+        public Result write(Object writable, Generator generator) {
```
jbaiera: It looks like it only supports Hadoop Writable objects here. For instance, Spark uses regular Java and Scala objects for data types. 

Since integrations can be added that have different types than described here, it makes me wonder if this is really the right place to have the connector ignore the document content for the bulk request. I don't think a ValueWriter will have enough wisdom at runtime to make the decision on a call-by-call basis. Maybe there needs to be work done in order to have a new `BulkCommand` that is returned from the `DeleteBulkFactory` ignore the document instead.

mr/src/main/java/org/elasticsearch/hadoop/cfg/Settings.java
```diff
@@ -221,6 +221,11 @@ public String getSerializerValueWriterClassName() {
         return getProperty(ES_SERIALIZATION_WRITER_VALUE_CLASS);
     }
 
+    public Settings setSerializerValueWriterClassName(String className) {
```
jbaiera: Since the ValueWriter doesn't seem like the right place to ignore the document source for delete operations, this won't be needed.

mr/src/main/java/org/elasticsearch/hadoop/rest/InitializationUtils.java
```diff
@@ -253,10 +253,18 @@ public static void validateSettings(Settings settings) {
             Assert.isTrue(settings.getMappingExcludes().isEmpty(), "When writing data as JSON, the field exclusion feature is ignored. This is most likely not what the user intended. Bailing out...");
         }
 
+        //check the configuration is coherent in order to use the delete operation
+        if (ConfigurationOptions.ES_OPERATION_DELETE.equals(settings.getOperation())) {
```
jbaiera: 👍 thanks for adding this

mr/src/main/java/org/elasticsearch/hadoop/rest/InitializationUtils.java
```diff
@@ -291,7 +299,8 @@ public static void validateSettingsForWriting(Settings settings) {
                 throw new EsHadoopIllegalArgumentException("Cannot use TTL on index/update requests in ES 6.x and " +
                         "above. Please remove the [" + ConfigurationOptions.ES_MAPPING_TTL + "] setting.");
             }
-        } else {
+        }
```
jbaiera: Nit: You can leave this formatting as is. Keeps the history cleaner, and the previous format is consistent with the project code style.

mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/DeleteBulkFactory.java
```diff
@@ -0,0 +1,99 @@
...
+            else if (writable instanceof MD5Hash) {
+                generator.writeString(writable.toString());
+            }
+            return Result.SUCCESFUL();
+        }
+    }
+
+    public DeleteBulkFactory(Settings settings, MetadataExtractor metaExtractor, EsMajorVersion version) {
+        // we only want a specific serializer for this particular bulk factory
+        super(settings.copy().setSerializerValueWriterClassName(NoDataWriter.class.getName()), metaExtractor, version);
```
jbaiera: Also not required if we use a new implementation of `BulkCommand`

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
```diff
@@ -64,43 +65,55 @@
...
-                { ConfigurationOptions.ES_OPERATION_CREATE, true, EsMajorVersion.V_7_X },
-                { ConfigurationOptions.ES_OPERATION_UPDATE, true, EsMajorVersion.V_7_X },
-                { ConfigurationOptions.ES_OPERATION_INDEX, false, EsMajorVersion.V_8_X },
-                { ConfigurationOptions.ES_OPERATION_CREATE, false, EsMajorVersion.V_8_X },
-                { ConfigurationOptions.ES_OPERATION_UPDATE, false, EsMajorVersion.V_8_X },
-                { ConfigurationOptions.ES_OPERATION_INDEX, true, EsMajorVersion.V_8_X },
-                { ConfigurationOptions.ES_OPERATION_CREATE, true, EsMajorVersion.V_8_X },
-                { ConfigurationOptions.ES_OPERATION_UPDATE, true, EsMajorVersion.V_8_X }
+        return Arrays.asList(new Object[][]{
+                {ConfigurationOptions.ES_OPERATION_INDEX, false, EsMajorVersion.V_1_X},
```
jbaiera: Nit: Thanks for adding these test cases, but can you revert the spacing changes on unaffected lines? Would help keep history tidy.

jbaiera: Or better yet, these could probably be generated with a couple loops.

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
```diff
@@ -258,18 +283,21 @@ public void testAll7X() throws Exception {
 
     @Test
     public void testIdPattern() throws Exception {
+        assumeFalse(isDeleteOP() && jsonInput);
         Settings settings = settings();
         if (version.onOrAfter(EsMajorVersion.V_8_X)) {
             settings.setResourceWrite("{n}");
-        } else {
+        }
```
jbaiera: Nit: revert newline

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
```diff
@@ -258,18 +283,21 @@ public void testAll7X() throws Exception {
...
+        else {
             settings.setResourceWrite("foo/{n}");
         }
 
         create(settings).write(data).copyTo(ba);
         String header;
         if (version.onOrAfter(EsMajorVersion.V_8_X)) {
             header = "{\"_index\":\"1\"" + (isUpdateOp() ? ",\"_id\":2" : "") + "}";
-        } else {
+        }
```
jbaiera: Nit: revert newline

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
```diff
@@ -523,16 +551,21 @@ private Settings settings() {
...
         InitializationUtils.setFieldExtractorIfNotSet(set, MapFieldExtractor.class, null);
         InitializationUtils.setBytesConverterIfNeeded(set, JdkBytesConverter.class, null);
         InitializationUtils.setUserProviderIfNotSet(set, HadoopUserProvider.class, null);
 
         set.setProperty(ConfigurationOptions.ES_WRITE_OPERATION, operation);
         if (version.onOrAfter(EsMajorVersion.V_8_X)) {
             set.setResourceWrite("foo");
-        } else {
+        }
```
jbaiera: Nit: revert newline

vincentarnaud90:
Hello,
Thanks for the reply. From what I undertstand from your comment, you ask me to do this. 
```java
	DeleteBulkFactory  extends AbstractBulkFactory {

		@Override
		public BulkCommand createBulk(){
			
			//Inside this BulkCommand, I can ignore the document source. 
			return DeleteTemplateBulk();
		}
	}
```

Instead of overriding the createBulk method in the DeleteBulkFactory class, can I just add my logic inside the createBulk method of AbstractBulkFactory class in order to return the DeleteTemplateBulk 
when the approriate settings are set ?
The High level code would look like this : 
```java
	public AbstractBulkFactory implements BulkFactory{
		@Override
		public BulkCommand createBulk(){
			//the code which is actually written

			if (isDeleteOp){
				//Inside this BulkCommand, I can ignore the document source. 
				return DeleteTemplateBulk();	
			}
			
		}	
	}
```

mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/DeleteBulkFactory.java
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.elasticsearch.hadoop.serialization.bulk;
+
+import org.apache.hadoop.io.*;
```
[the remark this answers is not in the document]
vincentarnaud90: Done

mr/src/main/java/org/elasticsearch/hadoop/rest/InitializationUtils.java
vincentarnaud90: Revert to one line

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
vincentarnaud90: I changed generation with loops. The array with boolean values is here to keep exactly the same order as before in the returned list

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
vincentarnaud90: Reverted

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
vincentarnaud90: Reverted

mr/src/test/java/org/elasticsearch/hadoop/serialization/CommandTest.java
vincentarnaud90: Reverted

jbaiera:
> Instead of overriding the createBulk method in the DeleteBulkFactory class, can I just add my logic inside the createBulk method of AbstractBulkFactory class in order to return the DeleteTemplateBulk
> when the approriate settings are set ?

I think that should be fine considering it already detects if the operation is an update vs an index operation.

vincentarnaud90:
Hello @jbaiera , sorry for the really late reply.
I have made the suggested changes. Can you have a look at them ?
Furthermore, it seems that your CI is not working anymore.

jbaiera:
Thanks for putting in those changes! I'll run the full test suite on my own since TravisCI is busted.

krish2code:
Hi @jbaiera, not sure if this is right place to post my question but its relevant to your closing comment. We are using ElasticSearch in AWS and are in need of Delete feature. As you mentioned above, this feature will be back ported in 7.8.0 but I don't see that the case. Delete is still an `Unknown operation` on 7.8 codeline. https://github.com/elastic/elasticsearch-hadoop/blob/7.8/mr/src/main/java/org/elasticsearch/hadoop/serialization/bulk/BulkCommands.java

krish2code:
@jbaiera As I mentioned above in my comment, we are using ElasticSearch on AWS and since latest version which is supported there is 7.4.2 and it does not have delete feature support. To mitigate that, I have created a PR against branch 7.4.2 to backport delete feature. Please review and let me know if I need to make any changes to code or process since this is my first commit here.  https://github.com/elastic/elasticsearch-hadoop/pull/1492

jbaiera:
@krish2code This backport fell between the cracks and ended up not making it in for the 7.8 release train. It should be available in 7.9. Sorry for the confusion!

satyaki-mallick:
Hi, is there a documentation on how to implement this?

sinban04:
@jbaiera

AFAIK, when i search the code in this repository (on the callpath of DataFrame sink), 
This PR is merged but actually it's not referenced by DataFrame Sink. 
I followed the code, 
and I found out that when EsDataFrameWriter (by EsRDDWriter) tries to send bulk data to ES Index,
It only uses `PUT` RESTful API 
```java
 Response response = execute(PUT, resource.bulk(), data);
```
https://github.com/elastic/elasticsearch-hadoop/blob/main/mr/src/main/java/org/elasticsearch/hadoop/rest/RestClient.java#L236

So far, i can tell delete operation is not yet provided on RDD nor DataFrame. Is it right ?
Is there any plan to support delete operation here ?

PR #1328, opened 2019-07-30
Remove LocalEs#setSingleNodeTemplate() method

martijnvg:
This is no longer need, because 1 shard is the default since 7.0


martijnvg:
I've updated the PR to only install the template if the cluster is on or before 6.x version.

PR #1330, opened 2019-08-02
Do not retry scroll requests in NetworkClient.

### LINKED ISSUE #1302: Timeouts on Scroll Requests can lead to dropped records
Scrolled searches are the primary avenue for retrieving data for jobs that read from Elasticsearch. Scroll sessions are started when a reader comes online, the scroll id from each request is used in the subsequent requests, and the session is closed when the scroll is exhausted.

When a node that is serving a scrolled request is overloaded, there is a decent chance that a client scroll request will time out waiting for the results. In these cases, ES-Hadoop will resubmit the scroll request to another node in the cluster. This resubmission can lead to issues if the original scroll request was being processed while the client timed out. 

The first scenario is that the client retries the request on another node and triggers a data race in Elasticsearch (https://github.com/elastic/elasticsearch/issues/35860). The server behaves as if two clients are accessing the scroll request at the same time, which is unsupported, even though it is the single client which has simply timed out waiting for the first request.

The second scenario is that the server successfully retrieves a page of scroll results and updates the internal scroll state to point to the next page of results, but the client times out waiting for the results to arrive and resubmits the request. The resubmitted request may end up succeeding, but the documents that were returned to the timed out request are essentially passed over and not retrieved. This can lead to a job reading fewer records than actually exist in Elasticsearch.

A work around to this issue is to increase the `es.http.timeout` property on ES-Hadoop to a longer time value, causing the connector to wait longer for the result of a search request instead of dropping the response and its contents due to a network timeout.

martijnvg:
Scroll requests cannot be retried on other nodes.
This can lead to less document being returned than is expected or
ES returning an error because a scroll id is accessed multiple times.

Closes #1302

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/NetworkClient.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/NetworkClient.java
@@ -20,8 +20,6 @@ package org.elasticsearch.hadoop.rest;
 
 import java.io.Closeable;
 import java.net.BindException;
-import java.net.NoRouteToHostException;
-import java.net.UnknownHostException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


ebuildy:
what about retrying on the same node ?

jbaiera:
@ebuildy Even retrying the same scroll operation on the same node can drop documents. If the node successfully returns results to the client, but the client has already timed out waiting for them, then those results are dropped. Scroll actions are stateful on the server side and a scroll id is not always unique for one batch of results.

ebuildy:
Thanks you, I understand! I was arrived here because got this issue ``failed (java.net.SocketException: Connection reset); Retrying has been disabled`` . I was using a K8S nginx ingress, moved to the Service directly fixed the issue.

PR #1333, opened 2019-08-16
Upgrade to and fix compatibility with Gradle 5.5

mark-vieira:
This pull request sorts out some issues related to upgrading to Gradle 5.5. We've replaced the way we implement "nested" variant builds with a simple `Exec` task instead of the `GradleBuild` task that has odd behavior due to it running in an embedded daemon.

There were some other misc things related to the upgrade and updated/remove APIs in Gradle 5.x.


mark-vieira:
Keep in mind that the main motivation here is so we can run this build with Java 11+. This will likely require some more tweaks to the build if the need to cross-compile sources against Java 8 remains.

Issue #1347, opened 2019-09-06
Repair the build's dependency on ES Build Tools


jbaiera:
For a few months now we have been executing builds against older versions of build tools from the Elasticsearch project. This issue is meant to capture the background and scope of the problem we are experiencing with our usage of build tools, as well as to detail the steps to resolve it.

Several months ago, some updates to build tools were made that raised the minimum supported version of Java. The build tools artifacts were targeting more recent versions of Java byte code than the ES-Hadoop build supports and builds were failing due to unsupported Java major minor failures. In order to increase the version of Java used to _compile_ ES-Hadoop we need to upgrade Gradle above 4.5.

Some efforts (#1253 #1255) were made to upgrade the project to newer versions of Gradle but they were eventually rolled back due to compatibility issues with the unified release process. ES-Hadoop uses nested builds to compile the Spark integration into variant artifacts for different Scala versions. These parts of the build began failing on Gradle 5.0+ due to some newly introduced locking code in core Gradle, and the upgrade was rolled back.

Moving forward, there are a few main steps to getting the build back on track. 

- [x] Fix the Scala variant compile process on Gradle 5.0+ (#1333)
- [x] Upgrade Gradle to 5.5 (#1333)
- [x] Deprecate and Remove support for Scala 2.10 in master (#1350)
- [x] Configure the build to use the same versions of Java for compilation as Elasticsearch (while still targeting Java 8 for compatibility) (#1351)
- [x] Return to using the appropriate versions of build tools for each development branch (#1351)

jbaiera:
Update here: I hit a snag while performing the upgrade to using Java 12 as the runtime for the Gradle build. It turns out that due to a bug in how Gradle's [Scala compiler merges fork options](https://github.com/gradle/gradle/blob/218902c09422799e4c20384c4c764a43025b1484/subprojects/language-scala/src/main/java/org/gradle/api/internal/tasks/scala/DaemonScalaCompiler.java#L86), fork options that have to do with which java executable to use when launching a task [are ignored](https://github.com/gradle/gradle/blob/218902c09422799e4c20384c4c764a43025b1484/subprojects/language-jvm/src/main/java/org/gradle/api/internal/tasks/compile/daemon/AbstractDaemonCompiler.java#L70-L78). This leads to Scala 2.10's Compiler to fail because it is being executed on Java 12, which is not supported. Switching to Java 11 experiences the same problem.

[...]

PR #1348, opened 2019-09-06
Set build to use external build tools during snapshots

jbaiera:
This PR sets the build to return to using non-local build tools dependencies when building snapshots.

Related to #1347


jbaiera:
Based on discussions offline, this will just be replaced with getting #1347 across the finish line

PR #1351, opened 2019-09-11
Upgrade build to use JDK 12 and current build tools

jbaiera:
This PR adds support to the build for running on JDK 12, while still compiling code to be compatible with Java 8. Additionally, since the version of Java has been updated, the usage of Elasticsearch's build-tools has been restored to the current 8.0 snapshot.

Relates #1347


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -64,6 +69,47 @@ class BuildPlugin implements Plugin<Project>  {
             println '==================================='
         }
     }
+
+    private static void configureBuildInfo(Project project) {
+        // Make sure the global build info plugin is applied to the root project first and foremost
+        // FIXHERE: Make a proper random test seed?
+        project.rootProject.ext.testSeed = "DEADBEEF"
```
mark-vieira: This should get moved into the `GlobalBuildInfoPlugin`. When I fix that I'll let you know to remove this.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -64,6 +69,47 @@ class BuildPlugin implements Plugin<Project>  {
...
+        // Hack new defaults into Global build info
+        if (!project.rootProject.ext.has('buildInfoConfigured')) {
+
+            JavaVersion minimumRuntimeVersion = JavaVersion.toVersion(Resources.getResourceContents("/minimumRuntimeVersion"))
+
+            // We snap the runtime to java 8 since Hadoop needs to see some significant
+            // upgrades to support any runtime higher than that
+            // Todo: Should this be provided via the ENV like the existing $RUNTIME_JAVA_HOME that we're overriding?
+            List<JavaHome> javaVersions = project.rootProject.ext.javaVersions as List<JavaHome>
+            JavaHome esHadoopRuntimeJava = javaVersions.find { it.version == 8 }
```
mark-vieira: FWIW, in the ES project we allow setting the `RUNTIME_JAVA_HOME` env var to do this. This might be preferable as you don't have to change the env var name itself as you bump stuff.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -64,6 +69,47 @@ class BuildPlugin implements Plugin<Project>  {
...
+                        '$JAVA8_HOME must be set to build ES-Hadoop. ' +
+                            "Note that if the variable was just set you might have to run `./gradlew --stop` for " +
+                            "it to be picked up. See https://github.com/elastic/elasticsearch/issues/31399 details."
+                )
+            }
+
+            // Set on global build info
+            GenerateGlobalBuildInfoTask generateTask = project.getTasks().getByName("generateGlobalBuildInfo") as GenerateGlobalBuildInfoTask
+            generateTask.setMinimumRuntimeVersion(minimumRuntimeVersion)
+            generateTask.setRuntimeJavaHome(esHadoopRuntimeJava.javaHome.getAbsoluteFile())
```
mark-vieira: If you end up going with the `RUNTIME_JAVA_HOME` method you can avoid the need to set this explicitly here.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -142,16 +188,18 @@ class BuildPlugin implements Plugin<Project>  {
...
 
-            File javaHome = findJavaHome()
-            // Register the currently running JVM version under its version number.
-            final Map<Integer, String> javaVersions = [:]
-            javaVersions.put(Integer.parseInt(JavaVersion.current().getMajorVersion()), javaHome)
-
-            project.rootProject.ext.javaHome = javaHome
-            project.rootProject.ext.runtimeJavaHome = javaHome
-            project.rootProject.ext.javaVersions = javaVersions
+            // FIXHERE: Still needed at all?
```
mark-vieira: Do we want to keep all this commented out stuff?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -314,6 +364,10 @@ class BuildPlugin implements Plugin<Project>  {
         compileTestJava.setSourceCompatibility('1.8')
         compileTestJava.setTargetCompatibility('1.8')
 
+        JavaCompile compileItestJava = project.tasks.getByName('compileItestJava') as JavaCompile
```
mark-vieira: We can set source and target compatibility on the project itself instead of doing it on each compile task individually.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/util/Resources.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.elasticsearch.hadoop.gradle.util;
```
mark-vieira: This should really go in `src/main/java`. Looks like the same exists for `WaitForURL.java`. Mixing .groovy and .java in `src/main/groovy` enables Groovy joint-compilation which incurs a compilation time penalty. If there's no need to have `.java` call out to `.groovy` then it's best to separate them. All your Groovy code will have access to your Java code.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Fri Sep 06 17:06:02 EDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-all.zip
```
dakrone: I think ES is on 5.6 now, should this upgrade to 5.6 also? (not as part of this PR, just in general)

Issue #1361, opened 2019-10-01
AbstractPigReadAsJsonTest fails multiple tests


dakrone:
From the test failure email, these tests failed:

```
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testFieldAlias[0]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testFieldAlias[1]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testFieldAlias[2]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTupleWithSchema[0]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTupleWithSchema[1]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTupleWithSchema[2]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTuple[0]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTuple[1]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTuple[2]
org.elasticsearch.hadoop.integration.pig.AbstractPigReadAsJsonTest testTuple[3]
```

All failed, however, none of the failures show up in the console from https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch-hadoop+master+periodic/123/console and there's no reproduction line.

jbaiera:
Fixed with #1364

PR #1364, opened 2019-10-11
Fix pig test failures

jbaiera:
In the master branch, Elasticsearch no longer returns the type metadata as part of a search operation. The Pig tests expect this data to be present and are thus failing. This PR fixes the logic in the pig tests to check the version before expecting the type to be present in metadata.


jbaiera:
Fixes #1361

Issue #1363, opened 2019-10-09
Spark javadoc is incomplete


jakelandis:
The `elasticsearch-spark-*` javadoc is missing some packages. For example https://www.javadoc.io/doc/org.elasticsearch/elasticsearch-spark-20_2.11/7.4.0 is missing `sql` and `streaming`. 

Cracking open the javadoc jar, it appears it is not being included in the jar. https://repo1.maven.org/maven2/org/elasticsearch/elasticsearch-spark-20_2.11/7.4.0/elasticsearch-spark-20_2.11-7.4.0-javadoc.jar

```
$ls org/elasticsearch/spark/
cfg rdd
```

jbaiera:
There is an additional problem here: Last time I checked, the Scaladoc and Javadoc tools do not play well together in terms of making a fully realized documentation artifact from a project that contains both Java and Scala sources. IIRC, we currently rely on only Javadoc outputs which does not include any documentation for Scala classes or packages.

I wonder if its worth it to explore creating both a scaladoc and javadoc output for the project.

jakelandis:
I started to play around with https://github.com/lightbend/genjavadoc ... so far it looks promising. Assuming I can figure out some gradle weirdness, I will submit a PR soon. 

An example screen shot:
<img width="1467" alt="image" src="https://user-images.githubusercontent.com/976291/66787284-b5970300-eea8-11e9-8e64-136a3aa6e2cc.png">

PR #1368, opened 2019-10-15
Add scala docs to javadocs

jakelandis:
This commit introduces a 3rd party plugin to co-locate
scala docs with java docs. https://github.com/lightbend/genjavadoc

Fixes: #1363

### DIFF
--- a/spark/sql-13/build.gradle
+++ b/spark/sql-13/build.gradle
@@ -124,7 +124,7 @@ jar {
 }
 
 javadoc {
-    dependsOn scaladoc //ensures that scala compiles and scala docs are valid
+    dependsOn compileScala
     source += project(":elasticsearch-hadoop-mr").sourceSets.main.allJava
     source += "$buildDir/generated/java"
     classpath += files(project(":elasticsearch-hadoop-mr").sourceSets.main.compileClasspath)
--- a/spark/sql-13/build.gradle
+++ b/spark/sql-13/build.gradle
@@ -12,6 +12,14 @@ variants {
     targetVersions '2.11.12'
 }
 
+configurations {
+    scalaCompilerPlugin {
+        defaultDependencies { dependencies ->
+            dependencies.add(project.dependencies.create( "com.typesafe.genjavadoc:genjavadoc-plugin_${scalaVersion}:0.13"))
+        }
+    }
+}
+
 println "Compiled using Scala ${project.ext.scalaMajorVersion} [${project.ext.scalaVersion}]"
 String sparkVersion = spark13Version
 
@@ -116,7 +124,9 @@ jar {
 }
 
 javadoc {
+    dependsOn scaladoc //ensures that scala compiles and scala docs are valid
     source += project(":elasticsearch-hadoop-mr").sourceSets.main.allJava
+    source += "$buildDir/generated/java"
     classpath += files(project(":elasticsearch-hadoop-mr").sourceSets.main.compileClasspath)
 }
 
@@ -141,4 +151,13 @@ configurations.all { Configuration conf ->
     }
 
     conf.exclude group: "org.mortbay.jetty"
+}
+
+tasks.withType(ScalaCompile) {
+    scalaCompileOptions.with {
+        additionalParameters = [
+                "-Xplugin:" + configurations.scalaCompilerPlugin.asPath,
+                "-P:genjavadoc:out=$buildDir/generated/java".toString()
+        ]
+    }
 }
--- a/spark/sql-20/build.gradle
+++ b/spark/sql-20/build.gradle
@@ -139,7 +139,7 @@ jar {
 }
 
 javadoc {
-    dependsOn scaladoc //ensures that scala compiles and scala docs are valid
+    dependsOn compileScala
     source += project(":elasticsearch-hadoop-mr").sourceSets.main.allJava
     source += "$buildDir/generated/java"
     classpath += files(project(":elasticsearch-hadoop-mr").sourceSets.main.compileClasspath)
--- a/spark/sql-20/build.gradle
+++ b/spark/sql-20/build.gradle
@@ -46,9 +46,7 @@ compileScala {
         "-Yno-adapted-args",
         "-Ywarn-dead-code",
         "-Ywarn-numeric-widen",
-        "-Xfatal-warnings",
-        "-Xplugin:" + configurations.scalaCompilerPlugin.asPath,
-        "-P:genjavadoc:out=$buildDir/generated/java".toString()
+        "-Xfatal-warnings"
     ]
 }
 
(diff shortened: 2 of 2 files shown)


jakelandis:
Use the following commands to test:

```
./gradlew elasticsearch-spark-20:javadoc
open spark/sql-20/build/docs/javadoc/index.html

./gradlew elasticsearch-spark-13:javadoc
open spark/sql-13/build/docs/javadoc/index.html
```

mark-vieira:
While reading [some of the documentation](https://github.com/lightbend/genjavadoc) I gotta say this sounds pretty gnarly. What's the main motivation of this over just using `scaladoc`? Certainly there's awkwardness here since we are presenting a Scala API as though it were a Java API? Not to mention we are adding to the already atrocious compile times of the Scala compiler by adding this Java source generation that is only ever used for JavaDoc generation but now runs on every compile.

This motivation also makes no sense to me:
> This project’s goal is the creation of real Javadoc for Scala projects. While Scaladoc—the native API documentation format of Scala—has several benefits over Javadoc, Java programmers are very much used to having access to API documentation in a syntax matching their programming language of choice.

But, this isn't Java code, it's Scala? So why are they doing all this hackiness to make a Scala API more amenable to Java programmers? I mean, aren't the consumers of this API Scala developers, and thereby more inclined to want the more robust Scaladoc vs a watered-down and potentially inaccurate Javadoc representation of the API?

mark-vieira:
To add, my comments are mainly rooting in my ignorance of the Scala ecosystem. If this is a common pattern for documentation of APIs implemented in Scala then by all means, I'll defer to the Scala experts here on whether this is a good idea. I simply found this method of JavaDoc generation to be a bit odd and want to ensure we have strong motivations for introducing this since it carries some cost.

spark/sql-13/build.gradle
```diff
@@ -116,7 +124,9 @@ jar {
 }
 
 javadoc {
+    dependsOn scaladoc //ensures that scala compiles and scala docs are valid
```
mark-vieira: We don't need to run `scaladoc` to generate the javadocs. We just need to _compile_. This should depend on `compileScala` instead.

jakelandis:
>  I mean, aren't the consumers of this API Scala developers

Nope. Any consumer that wants to use this library with Spark will need to use this API. That could Java or Scala devs. My guess is most are Java devs. If we think most are Scala devs we should look at publishing the scala docs jar. 

> If this is a common pattern for documentation of APIs implemented in Scala then by all means

Not sure if this is a common pattern, but the library is by from [Lightbend](https://en.wikipedia.org/wiki/Lightbend)  formerly known as Typesafe, which is well known.

spark/sql-13/build.gradle
jakelandis: since nothing else actually runs the scala docs, I thought that generating them next to the java docs (even if not used) would serve as a good as test. But it is unnecessary and I will change it.

mark-vieira:
> Nope. Any consumer that wants to use this library with Spark will need to use this API. That could Java or Scala devs. My guess is most are Java devs. If we think most are Scala devs we should look at publishing the scala docs jar.

Gotcha, understood. Yeah, in that case I would agree then that having the API in Javadoc format would be most accessible. I assume while you were testing this the compile-time cost wasn't appreciable? If it is, there are perhaps things we could do, like make the compiler plugin conditional on the `javadoc` task being part of the task graph.

jbaiera:
> Nope. Any consumer that wants to use this library with Spark will need to use this API. That could Java or Scala devs. My guess is most are Java devs. If we think most are Scala devs we should look at publishing the scala docs jar.

We should probably avoid making these assumptions without any data. My gut feeling is that it's split 40/40/20 across Scala, Python, and Java (respectively), but I don't have any data to back this up other than past issues I can think of.

> While reading some of the documentation I gotta say this sounds pretty gnarly.

From what I'm reading, it "emits structurally equivalent Java code for all Scala sources of a project, keeping the Scaladoc comments (with a few format adaptions)." Is this structural Java code created alongside the regular compiler class file output, or is this structural Java code an intermediary step? I'm assuming its the former, but I want us to be sure. Scala has some whacky language features in it that just do not translate well to structural Java code.

jbaiera:
> Nope. Any consumer that wants to use this library with Spark will need to use this API. That could Java or Scala devs. 

Is the Java API documentation missing because it's nested under the Scala source? I want to make sure I'm understanding this clearly since I admittedly haven't spent much time looking into the Javadoc/Scaladoc problem for this part of the project.

jakelandis:
>  I assume while you were testing this the compile-time cost wasn't appreciable? 

The javadoc task doesn't get called until 'pack' , which should only happen during the release, and doesn't add any appreciable time. I also checked `compileScala` in case the new params impacted it in a funny way, which it does not.  

> Is this structural Java code created alongside the regular compiler class file output, or is this structural Java code an intermediary step? 

I'm not sure I completely follow the question, but the plugin puts the generated java code in a defined directory. 
```
$ls spark/sql-20/build/classes
scala

$ls spark/sql-20/build/generated
java    sources
```
Just to be certain, I diffed the generated artifacts and the only difference is the addition of the scala javadoc. (the left is with the changes here)
image

> Is the Java API documentation missing because it's nested under the Scala source? 

Where is the Java API documentation missing ? 

> My gut feeling is that it's split 40/40/20 across Scala, Python, and Java (respectively)

We should probably consider adding the proper scala docs jar to the things that published ?  However, I don't think this PR and that are mutually exclusive things.

mark-vieira:
> The javadoc task doesn't get called until 'pack' , which should only happen during the release, and doesn't add any appreciable time. I also checked compileScala in case the new params impacted it in a funny way, which it does not.

Right, since this is implemented as a Scala compiler plugin those Java source files are generated whenever the Scala source is compiled, regardless of whether we're actually generated JavaDocs or not, so that's the bit I would expect to have a cost. I suspect one of the most costly parts of the Scala compile process is building the AST, which needs to be done regardless of this generation step so in way it makes sense that it doesn't add much time to the build. 👍

jbaiera:
> I'm not sure I completely follow the question, but the plugin puts the generated java code in a defined directory.

Ok cool, this was my major paranoia. I was mostly worried that the compiler plugin was taking the scala source, turning it into java source, and then converting the resulting java source into byte code/class files. But it sounds like the scala source still compiles to byte code and class files directly, with the added generation of some Java files for use in the javadoc process.

spark/sql-20/build.gradle
```diff
@@ -38,7 +46,9 @@ compileScala {
         "-Yno-adapted-args",
         "-Ywarn-dead-code",
         "-Ywarn-numeric-widen",
-        "-Xfatal-warnings"
+        "-Xfatal-warnings",
+        "-Xplugin:" + configurations.scalaCompilerPlugin.asPath,
```
jbaiera: I see that in the sql-13 build file we add these with `tasks.withType(...)` instead of appending here. I think I'm fine with either, but we should try to do the same in both files.

jakelandis: done https://github.com/elastic/elasticsearch-hadoop/pull/1368/commits/2e42e73253e919392b78dfc8bb8c17769ff48d16

jakelandis:
@jbaiera - should be ready for another look. thanks!

PR #1369, opened 2019-10-16
Fix spark test failures

jbaiera:
Recently the `_source` endpoints have had their types removed. There are a few spots in the test code that uses these end points that really don't need to be using them in the first place. This PR removes them to get the tests passing again.


colings86:
@jbaiera Please remember to add version labels to all PRs

PR #1371, opened 2019-10-18
Fix CI failures due to Hadoop Distro

jbaiera:
CI jobs run with the `-Pdistro=hadoopYarn` setting, which changes the Hadoop dependency at runtime to 2.7 instead of 2.2. This causes the CI jobs to fail because the SHA files for the dependency to not match the new dependencies. This PR switches the default hadoop distro used for building to be consistent with the one used in CI.


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -171,8 +171,7 @@ class BuildPlugin implements Plugin<Project>  {
                     project.rootProject.ext.hadoopClient = ["org.apache.hadoop:hadoop-client:$version"]
                     println "Using Apache Hadoop on YARN [$version]"
                     break
-
-                default:
+                case "hadoopStable":
```
dakrone: We should add a `default:` case and throw an error, so we don't accidentally introduce leniency in the future

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -171,8 +171,7 @@ class BuildPlugin implements Plugin<Project>  {
...
                     String version = project.hadoop22Version
                     project.rootProject.ext.hadoopVersion = version
                     project.rootProject.ext.hadoopClient = ["org.apache.hadoop:hadoop-client:$version"]
```
dakrone: There's no difference in the code of the two different switch paths other than the `println`, can be move the settings out? Unless we need to *not* set them in the 'default' case (in which case we should be explicit in that)

PR #1387, opened 2019-11-08
Update text around es.read.field.as.array.include

jbaiera:
This setting is often misunderstood and misused, especially when it comes to nested fields. This PR updates the docs in an attempt to better explain the usage of the field with more information in the examples.

### DIFF
--- a/docs/src/reference/asciidoc/core/configuration.adoc
+++ b/docs/src/reference/asciidoc/core/configuration.adoc
@@ -277,7 +277,7 @@ es.read.field.exclude = company.income
 added[2.2]
 `es.read.field.as.array.include` (default empty)::
 Fields/properties that should be considered as arrays/lists. Since {es} can map one or multiple values to a field, {eh} cannot determine from the mapping
-whether to treat a field on a document as a single value or an array of values. When encountering multiple values, {eh} will automatically read
+whether to treat a field on a document as a single value or an array. When encountering multiple values, {eh} automatically reads
 the field into the appropriate array/list type for an integration, but in strict mapping scenarios (like Spark SQL) this
 may lead to problems (an array is encountered when Spark's Catalyst engine expects a single value).
 The syntax for this setting is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude].
--- a/docs/src/reference/asciidoc/core/configuration.adoc
+++ b/docs/src/reference/asciidoc/core/configuration.adoc
@@ -283,7 +283,7 @@ may lead to problems (an array is encountered when Spark's Catalyst engine expec
 The syntax for this setting is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude].
 Multiple values can be specified by using a comma. By default, no value is specified, meaning no properties/fields are treated as arrays.
 
-NOTE: Not all fields need to specify `es.read.field.as.array` in order to be treated as an array in {eh}. Fields of type `nested` are always
+NOTE: Not all fields need to specify `es.read.field.as.array` to be treated as an array. Fields of type `nested` are always
 treated as an array of objects, and should not be marked under `es.read.field.as.array.include`.
 
 For example, given the document:
--- a/docs/src/reference/asciidoc/core/configuration.adoc
+++ b/docs/src/reference/asciidoc/core/configuration.adoc
@@ -284,7 +284,7 @@ The syntax for this setting is similar to that of {es} {ref}/search-request-body
 Multiple values can be specified by using a comma. By default, no value is specified, meaning no properties/fields are treated as arrays.
 
 NOTE: Not all fields need to specify `es.read.field.as.array` to be treated as an array. Fields of type `nested` are always
-treated as an array of objects, and should not be marked under `es.read.field.as.array.include`.
+treated as an array of objects and should not be marked under `es.read.field.as.array.include`.
 
 For example, given the document:
 [source,json]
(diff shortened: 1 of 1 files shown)


docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
 added[2.2]
 `es.read.field.as.array.include` (default empty)::
 Fields/properties that should be considered as arrays/lists. Since {es} can map one or multiple values to a field, {eh} cannot determine from the mapping
-whether to instantiate one value or a array type (depending on the library type). When encountering multiple values, {eh} will automatically use 
-the array/list type but in strict mapping scenarios (like Spark SQL) this might lead to an unexpected schema change.
-The syntax is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude]. 
-Multiple values can be specified by using a comma. By default, no value is specified meaning no properties/fields are included.
+whether to treat a field on a document as a single value or an array of values. When encountering multiple values, {eh} will automatically read
```
jrodewig:
```suggestion
whether to treat a field on a document as a single value or an array. When encountering multiple values, {eh} automatically reads
```

docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
...
+the field into the appropriate array/list type for an integration, but in strict mapping scenarios (like Spark SQL) this
+may lead to problems (an array is encountered when Spark's Catalyst engine expects a single value).
+The syntax for this setting is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude].
+Multiple values can be specified by using a comma. By default, no value is specified, meaning no properties/fields are treated as arrays.
 
-For example:
+NOTE: Not all fields need to specify `es.read.field.as.array` in order to be treated as an array in {eh}. Fields of type `nested` are always
```
jrodewig:
```suggestion
NOTE: Not all fields need to specify `es.read.field.as.array` to be treated as an array. Fields of type `nested` are always
```

docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
...
+treated as an array of objects, and should not be marked under `es.read.field.as.array.include`.
```
jrodewig:
```suggestion
treated as an array of objects and should not be marked under `es.read.field.as.array.include`.
```

docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
...
+[source,json]
+----
+{
+  "foo": {
+    "bar": [[["abc", "def"], ["ghi", "jkl"]], [["mno", "pqr"], ["stu", "vwx"]]]
+  }
+}
+----
+
+Then specify the dimensionality of the array in the configuration
```
jrodewig:
```suggestion
Then specify the dimensionality of the array in the configuration:
```

docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
...
-whether to instantiate one value or a array type (depending on the library type). When encountering multiple values, {eh} will automatically use 
-the array/list type but in strict mapping scenarios (like Spark SQL) this might lead to an unexpected schema change.
-The syntax is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude]. 
-Multiple values can be specified by using a comma. By default, no value is specified meaning no properties/fields are included.
+whether to treat a field on a document as a single value or an array of values. When encountering multiple values, {eh} will automatically read
+the field into the appropriate array/list type for an integration, but in strict mapping scenarios (like Spark SQL) this
+may lead to problems (an array is encountered when Spark's Catalyst engine expects a single value).
+The syntax for this setting is similar to that of {es} {ref}/search-request-body.html#request-body-search-source-filtering[include/exclude].
+Multiple values can be specified by using a comma. By default, no value is specified, meaning no properties/fields are treated as arrays.
```
jrodewig: Makes this consistent with the def's first sentence.
```suggestion
Multiple values can be specified by using a comma. By default, no value is specified, meaning no fields/properties are treated as arrays.
```

docs/src/reference/asciidoc/core/configuration.adoc
```diff
@@ -277,19 +277,47 @@ es.read.field.exclude = company.income
...
+  "foo": {
+    "bar": [[["abc", "def"], ["ghi", "jkl"]], [["mno", "pqr"], ["stu", "vwx"]]]
+  }
+}
+----
+
+Then specify the dimensionality of the array in the configuration
+[source,ini]
+----
+# mapping foo.bar as a 3-level/dimensional array
```
jakelandis: should this also have a note - this is not needed for nested types since they are always mapped to arrays ? (It is unclear if multi-dimensional arrays still need to be declared for nested types)

jbaiera: After thinking on this, yes - multidimensional nested fields will need to be mapped, but will need to use a number of dimensions that is one fewer than actually exist. This feels really counter intuitive, especially with how complicated these settings already are.

I might open an issue for reconciling these problems. Nested fields should be detected as array fields, but if they are listed on the array includes, we should defer to that value as the number of nested levels for the arrays.

jakelandis: I'll defer to your judgement around the advice given/not-given here.

PR #1388, opened 2019-11-11
Adapt to upstream for type safe build globals

jasontedor:
This commit adapts the elasticsearch-hadoop build to an upstream change to make some build globals type safe. In particular, we adapt the usage of Java versions.

Relates elastic/elasticsearch#48778

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/InstanceInfo.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/InstanceInfo.groovy
@@ -20,6 +20,7 @@
 package org.elasticsearch.hadoop.gradle.fixture.hadoop
 
 import org.apache.tools.ant.taskdefs.condition.Os
+import org.elasticsearch.gradle.info.BuildParams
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/InstanceInfo.groovy
```diff
@@ -236,7 +237,7 @@ class InstanceInfo {
 
     /** Return the java home used by this node. */
     String getJavaHome() {
-        return javaVersion == null ? project.runtimeJavaHome : project.javaVersions.find { it.version == javaVersion }.javaHome.absolutePath
+        return javaVersion == null ? project.runtimeJavaHome : BuildParams.javaVersions.find { it.version == javaVersion }.javaHome.absolutePath
```
mark-vieira: Assuming we refactor out usages of extra properties completely, we can replace `project.runtimeJavaHome` with the `BuildParams` variant.

jasontedor:
@mark-vieira Yes, this is enough to make the build successful. I would be happy if you could pick up the follow-ups.

PR #1399, opened 2019-12-04
Upgrade to Gradle 6

jbaiera:
Upgraded to Gradle 6 and tested against a snapshot of the updated Elasticsearch build tools. This PR also has some small changes to simplify testing with the local snapshot directory.


buildSrc/build.gradle
```diff
@@ -78,6 +78,9 @@ dependencies {
 
     if (localRepo) {
         compile name: "build-tools-${buildToolsVersion}"
+        // Required for dependency licenses task (explicitly added in case of localRepo missing transitive dependencies)
```
mark-vieira: I'm confused why this is necessary. Is the POM we publish to the local repo not including these dependencies?

jbaiera: The POM isn't copied over into the local repo directory at the moment in any automation. If it were, would Gradle be able to pick these dependencies up?

mark-vieira: Perhaps, depends on how we are registering the repo. If we are using a flat dir repo then, no, as it doesn't handle metadata.

Issue #1400, opened 2019-12-05
[FEATURE] es.batch.write.wait - time to wait between each bulk


roeyshemtov:
### What kind an issue is this?

-  Feature Request. 

### Feature description
I am running a scenario that i have a big a dataset that i want to index into elastic, but i can't controll the index rate without changes the number of partitions of the dataset(that mean lower/higher parallel writing).
I know there is bulk size and bulk entries but it just adjust the bulking.
what im suggest is to add a configurable es.batch.write.wait, and that is the time in ms that it`ll wait between each bulk.

This configuration should help to adjust too fast index rate without chaning the dataset itself.

What do you think, it will be helpful for other users? because this feature will be great at our usecase.

jbaiera:
Sorry for the long wait on a response here. I think this is an interesting idea. Many parts of Elasticsearch's internal components have throttling to adjust how much work they're performing to avoid impact on the system, but the throttling that Elasticsearch implements is to avoid internal contention for resources on the Elasticsearch nodes themselves. Another place I can find throttling logic is in Logstash, though it seems that feature is more of a traffic control filter meant to cap the throughput of data that you expect to be within a certain number of events per time unit.

My main concern here is that while introducing throttling options would be a straightforward enhancement, is it the correct action to take? Throttling a job would indicate (to me) that Elasticsearch is not effectively handling the ingestion load, which I feel is a more pressing concern to address in terms of scalability.

Obviously, everyone's use case is unique, so I'm interested to hear what kind of issues you've run into, and how adding a throttle here would assist in fixing those issues vs. improving the general throughput of the system.

roeyshemtov:
Hello @jbaiera and thanks for answer.
I am partly agree that the infrastructure need to handle it, but in our case we can`t improve the general throughput of the system, and we don`t care that our Spark ETLs will take longer.
We are two teams that using elasticsearch 5 and we facing two problems that this issue can fix:
1. My problem - we have an infrastructure team (elasticsearch environment admins), that doesn`t allow us to pass a MAX limit of index rate of the environment - so here it can help.
2. Copartenr problems - They running a little cluster of elasticsearch by their own, and when they write their data (~2B rows) the elasticsearch cluster is just crash. (They doesn`t care the ETL will take longer and they can`t increase their environment - they have enough storage for the data).

masseyke:
[...]

PR #1405, opened 2019-12-11
[1400] Added es.batch.write.wait - time to wait between each bulk

roeyshemtov:
According to issue:
#1400


mr/src/main/java/org/elasticsearch/hadoop/rest/bulk/BulkProcessor.java
```diff
@@ -210,6 +212,16 @@ public BulkResponse tryFlush() {
                     // Log messages, and if wait time is set, perform the thread sleep.
                     initFlushOperation(bulkLoggingID, retryOperation, retries.size(), waitTime);
 
+                    // Check if should wait before sending the request.
+                    try {
+                        if (this.batchWriteWaitTime != 0 && !retryOperation)
+                            Thread.sleep(batchWriteWaitTime);
+                    }
```
jbaiera: I think in this case we would want something more robust to measure the time between bulk operations. If the wait time is 30 seconds but it takes 20 seconds to generate enough records to perform a bulk operation, then the total time between bulks is 50 seconds instead of 30. Since the time spent reaching the batch size limit is variable, the time to wait between each bulk operation is variable as well.

Additionally, the bulk operation does not have a consistent size in bytes or number - rather, the bulk operation is executed based on which ever limit is hit first. This variability of batch size affects how long it takes for ES to persist the data. When a batch operation would be executed and with how many records is completely disconnected from the wait time, which means that this property can become a moving target in terms of getting the right value.

I think instead of that, we would want to measure and cap the rate of ingestion to a configured "event per time unit" target rate. This is similar to how Elasticsearch's Reindex operation does throttling (https://www.elastic.co/guide/en/elasticsearch/reference/7.5/docs-reindex.html#docs-reindex-api-query-params). This is much easier to reason about and does a more consistent job of cohesively managing the rate of indexing than just a wait time.

roeyshemtov: I agree it is not good solution to wait between bulks, but i don`t know if we can implement "event per time unit" because it is a little bit fake ( i mean we have x Executors that each executor will write in that index rate).
How about starting a timer after each bulk finish and before the next bulk writing wait till timer finish?

jbaiera:
Closing this out since we're not convinced that this is the correct approach to managing ingestion rate.

PR #1408, opened 2019-12-12
Docs: Remove a broken [float]

nik9000:
This float tag is broken and docbook has been ignoring it.
`--direct_html` doesn't ignore it though.


nik9000:
thanks @jbaiera !

PR #1413, opened 2020-01-11
support es shard number as partition number

saiergong:
This commit can change the execution time from hours to minutes in some scenarios like：
1) query a lot of index and each index has one shard, like thousands of index
2) each shard has ten millions of docs, and aboud 20GB size
3) user use match phrase query, the term will match a lot but the phrase match little

in scenarios like above, when we count partitions, we will query each shard to count hit docs, because we use match phrase, each shard will spend about one second average, then it will spend about hours in `findSlicePartitions` function total.

so maybe we shoud support parameter to skip this count if each shard hit little.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/cfg/ConfigurationOptions.java
@@ -142,6 +142,9 @@ public interface ConfigurationOptions {
     String ES_INPUT_JSON = "es.input.json";
     String ES_INPUT_JSON_DEFAULT = "no";
 
+    String ES_INPUT_SHARDS_AS_PARTITION = "es.input.shard.num.as.partition.num";
+    String ES_INPUT_SHARDS_AS_PARTITION_DEFAULT = "no";
+
     /** Index settings */
     String ES_INDEX_AUTO_CREATE = "es.index.auto.create";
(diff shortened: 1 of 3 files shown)


jbaiera:
I don't think we should be adding yet another setting for modifying the logic of selecting partition counts. We already have `es.input.max.docs.per.partition`, which if it is set will cause the connector to perform the counts per shard.

As it stands, there is no longer a default value for `es.input.max.docs.per.partition` in the 7.x line, which should have the connector default to using the number of shards as the number of partitions without doing any shard counting.

Are you setting `es.input.max.docs.per.partition` at the moment? Does the performance problem still exist when that setting is removed?

saiergong:
i'm using elasticsearch hadoop 5.2.0, even i do not set `es.input.max.docs.per.partition`, the performance problem still exist. because the 5.2.0 version will always  perform the counts per shard request.

jbaiera:
Unfortunately the 5.x line is no longer under development and won't see any more releases going forward. Even if this PR were to go through, you would need to update to a newer version (a version on 7.x at the moment) to see the changes. 

As mentioned before, on the newer versions the code that performs the shard counts is disabled by default with the expectation that users can turn the slicing feature on if they have need of it, which should be rare.

Because of this I'm going to go ahead and close this PR. We can reopen it if there is a compelling enough reason to, but I would suggest upgrading first since this issue has already been addressed.

PR #1425, opened 2020-02-11
Bump version to 7.6.1

polyfractal:
Note: I did not update the README since it already points to 7.6.0 for the stable release, which I believe is correct?


jakelandis:
LGTM

jbaiera:
@polyfractal That is correct, thank you!

PR #1430, opened 2020-02-24
Update the QA project to use test cluster api

jbaiera:
This PR includes changes necessary to run the Kerberos QA tests using the new test clusters functionality in build-tools. 

This also updates the project to run fixtures on the Java 8 runtime. JAVA_HOME can now be set on cluster configs and is used by fixtures.

A cluster can be configured with a dependent ElasticsearchCluster. This cluster is depended on for any configuration tasks in case its cluster address is used.

Settings are now stored as objects in cluster configurations instead of Strings since they may be GStrings with closures inside of them that must be resolved ONLY at configuration writing time. This is needed to resolve test cluster http addresses which are only discoverable once the cluster is started.

In order to make this resolution process simpler, removed most of the Map<String, String> types in the code and replaced them with the FileSettings type which can resolve closures as needed when writing the configuration to a file.

Upgraded Spark QA runtime to 2.3.4 as the older version has been removed from apache's mirrors.

AbstractClusterTask now extends DefaultTestClustersTask, as it is the only way to ensure that Elasticsearch is still running when starting the task.

In order to configure JAVA_HOME consistently, each AbstractClusterTask now has its environment variables resolved in its super class instead of with duplicated code in each task.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopClusterFormationTasks.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopClusterFormationTasks.groovy
@@ -23,6 +23,7 @@ import org.apache.tools.ant.DefaultLogger
 import org.elasticsearch.gradle.LoggedExec
 import org.elasticsearch.gradle.Version
 import org.elasticsearch.gradle.test.Fixture
+import org.elasticsearch.gradle.testclusters.DefaultTestClustersTask
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfiguration
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.InstanceConfiguration
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.RoleConfiguration
@@ -42,6 +43,8 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 import java.nio.file.Paths
 
+import static org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.SettingsContainer.FileSettings
+
 /**
  * A helper for creating tasks to build a cluster that is used by a task, and tear down the cluster
  * when the task is finished.
@@ -49,11 +52,12 @@ import java.nio.file.Paths
 class HadoopClusterFormationTasks {
 
     /**
-     * A start and stop task for a fixture
+     * A start and stop task for an instance, as well as any and all instance specific setup and teardown tasks
      */
-    static class TaskPair {
+    static class InstanceTasks {
         Task startTask
         Task stopTask
+        List<Task> allTasks
     }
 
     /**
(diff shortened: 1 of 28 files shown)


qa/kerberos/build.gradle
```diff
@@ -84,7 +84,7 @@ dependencies {
 
 // Disable the integration tests for Kerberos until we can find a solution to the failures due to + sign 
 // in the file path on CI.
-boolean disableTests = true
+boolean disableTests = false
```
jakelandis: can we remove this boolean and comment ?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopClusterFormationTasks.groovy
```diff
@@ -168,6 +172,16 @@ class HadoopClusterFormationTasks {
                             }
                         }
                     }
+
+                    // Check to see if any of the instance tasks are test cluster aware, and if they are, set the
+                    // es cluster to be whichever cluster was configured, if any
+                    if (instanceInfo.elasticsearchCluster != null) {
```
jakelandis: As discussed in person, if this is unnecessary and you can configure it in-place, I think that would avoid a possible point of confusion.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopClusterFormationTasks.groovy
```diff
@@ -42,18 +43,21 @@ import org.apache.tools.ant.taskdefs.condition.Os
...
 
     /**
-     * A start and stop task for a fixture
+     * A start and stop task for an instance, as well as any and all instance specific setup and teardown tasks
      */
-    static class TaskPair {
+    static class InstanceTasks {
         Task startTask
         Task stopTask
+        List<Task> allTasks
```
jakelandis: As discussed in person, the allTasks is a bit confusing.

qa/kerberos/build.gradle
jbaiera: Actually this should stay disabled for the time being since it is still unlikely to pass general CI without some other changes.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopClusterFormationTasks.groovy
```diff
@@ -69,7 +72,7 @@ class HadoopClusterFormationTasks {
      * <p>
      * Returns a list of NodeInfo objects for each node in the cluster.
      *
-     * Based on {@link org.elasticsearch.gradle.test.ClusterFormationTasks}
+     * Based on (now removed) org.elasticsearch.gradle.test.ClusterFormationTasks
```
mark-vieira: Is this really necessary then?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/util/ObjectUtil.groovy
```diff
@@ -0,0 +1,32 @@
...
+ * under the License.
+ */
+
+package org.elasticsearch.hadoop.gradle.util
+
+class ObjectUtil {
+    static String unapplyString(Object value) {
+        if (value == null) {
+            return null
+        } else if (value instanceof Closure) {
```
mark-vieira: You can make this even more generic and just use `Callable`.

jbaiera:
Relates #1423

PR #1438, opened 2020-03-09
Replace apache download tasks with custom ivy repository.

jbaiera:
Hadoop ecosystem tarballs are hosted on Apache's download mirrors only instead of in maven repositories. Previously we were downloading these artifacts from a round robin of mirror endpoints, using securely obtained hashes to validate the downloaded files even from non HTTPS sources.

This PR removes those download tasks, replacing it with a custom Ivy repository backed by an HTTPS enabled apache mirror. This allows us to ensure the artifacts are securely downloaded and cached by Gradle instead of relying on task outputs like the old mirror download task.


mark-vieira:
One thing to keep in mind here is that Gradle doesn't do artifact checksum verification by default. In fact, this capability was only recently added in Gradle 6.2. As it is, this PR _removes_ the checksum validation for these artifacts. If we want to retain that capability, we'll need to enabled checksum verification in Gradle [as described here](https://docs.gradle.org/current/userguide/dependency_verification.html#sec:checksum-verification).

jbaiera:
The dependency verification features look really useful, but I think they might be a bit out of scope. It looks like we mostly want the verification for downloads from apache mirrors, but it's a global setting that will require upkeep outside of normal project configuration. Additionally, we would need to maintain a number of unverified configurations since we rely on the local file repo in snapshot and release builds.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/HadoopFixturePlugin.groovy
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    private static configureApacheMirrorRepository(Project project) {
+        RepositoryHandler repositoryHandler = project.getRepositories()
+        repositoryHandler.add(repositoryHandler.ivy(new Action<IvyArtifactRepository>() {
+            @Override
+            void execute(IvyArtifactRepository ivyArtifactRepository) {
+                ivyArtifactRepository.setUrl(APACHE_MIRROR)
+                ivyArtifactRepository.patternLayout(new Action<IvyPatternRepositoryLayout>() {
```
mark-vieira: Why aren't we using a lambda for all these `Action<T>` implementations?

PR #1439, opened 2020-03-09
Remove root project testing jar

jbaiera:
ES-Hadoop currently relies on its integration test code to be jar'd up before the tests are executed since many of the integrations rely on a jar being provided in order to run test code. We have a decent amount of shared code in the integration test projects, so we package it all up into one big jar on the root project, which makes configuring tests very confusing.

This PR removes that root testing jar and instead builds an itest jar in each project that needs it. We still have an issue with the shared code between integration tests, but there will be a follow up PR to address that.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -583,15 +583,15 @@ class BuildPlugin implements Plugin<Project>  {
      */
     private static void configureIntegrationTestTask(Project project) {
         if (project != project.rootProject) {
-            Jar itestJar = project.tasks.create('itestJar', Jar)
-            itestJar.dependsOn(project.tasks.getByName('jar'))
-            itestJar.getArchiveClassifier().set('testing')
-            project.logger.info("Created [${project.name}] Testing Jar")
+            TaskProvider<Task> itestJar = project.tasks.register('itestJar', Jar) { Jar itestJar ->
+                itestJar.dependsOn(project.tasks.getByName('jar'))
+                itestJar.getArchiveClassifier().set('testing')
 
-            // Add this project's classes to the testing uber-jar
-            itestJar.from(project.sourceSets.main.output)
-            itestJar.from(project.sourceSets.test.output)
-            itestJar.from(project.sourceSets.itest.output)
+                // Add this project's classes to the testing uber-jar
+                itestJar.from(project.sourceSets.main.output)
+                itestJar.from(project.sourceSets.test.output)
+                itestJar.from(project.sourceSets.itest.output)
+            }
 
             Test integrationTest = project.tasks.create('integrationTest', RestTestRunnerTask.class)
             integrationTest.dependsOn(itestJar)
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -37,7 +37,6 @@ import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.api.tasks.javadoc.Javadoc
 import org.gradle.api.tasks.testing.Test
-import org.gradle.api.tasks.testing.TestReport
 import org.gradle.external.javadoc.JavadocOutputLevel
 import org.gradle.external.javadoc.MinimalJavadocOptions
 import org.gradle.internal.jvm.Jvm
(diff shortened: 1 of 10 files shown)


hive/build.gradle
```diff
@@ -21,6 +21,14 @@ jar {
     }
 }
 
+itestJar {
+    from(zipTree(project(":elasticsearch-hadoop-mr").jar.archivePath)) {
```
jakelandis: I think the preferred way to model this is to via configuration/artifacts/dependencies instead of referencing the project directly. 

So ... this build.gradle would looks similar to
```
configurations {
    sharedHadoop
    sharedSpi
}
...
itestJar {
    from configurations.sharedHadoop
    from configurations.sharedSpi
}
...
dependencies {
 ...
    sharedHadoop project(path: ':elasticsearch-hadoop-mr', configuration: 'sharedHadoop')
    sharedSpi project(path: ':elasticsearch-hadoop-mr', configuration: 'sharedSpi')
}
```
and mr/build.gradle would look something like
```
configurations {
    sharedHadoop
    sharedSpi
}

artifacts{
    sharedHadoop(new File(projectDir, "src/main/java/org/elasticsearch/hadoop"))
    sharedSpi(new File(projectDir, "src/main/resources/META-INF/services"))
}
```

(and don't have an answer for the generated esh-build.properties)

However, implementing something like that is probably outside the scope of this PR especially since what you have here matches the existing model.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -590,72 +582,52 @@ class BuildPlugin implements Plugin<Project>  {
...
-            showCauses true
-            showExceptions true
-            showStackTraces true
-            stackTraceFilters "groovy"
-            minGranularity 2
-            maxGranularity 2
-        }
+        if (project != project.rootProject) {
+            Jar itestJar = project.tasks.create('itestJar', Jar)
+            itestJar.dependsOn(project.tasks.getByName('jar'))
```
mark-vieira: Why do we need the regular JAR? We are just packaging the sourceset output, I don't actually need to bundle the normal JAR to create this one.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -590,72 +582,52 @@ class BuildPlugin implements Plugin<Project>  {
...
+            itestJar.getArchiveClassifier().set('testing')
+            project.logger.info("Created [${project.name}] Testing Jar")
```
mark-vieira: FYI, this is being logged at _configuration time_ as is. If we want to add this logging when we _actually_ build the jar we need to place it in a task action like `doLast`.

mr/src/itest/java/org/elasticsearch/hadoop/Provisioner.java
```diff
@@ -42,7 +42,7 @@
         // init ES-Hadoop JAR
         // expect the jar under build\libs
         try {
-            File folder = new File(".." + File.separator + "build" + File.separator + "libs" + File.separator).getCanonicalFile();
+            File folder = new File("build" + File.separator + "libs" + File.separator).getCanonicalFile();
```
mark-vieira: Let's avoid hardcoding "build" and instead use `project.buildDir`.

hive/build.gradle
jbaiera: This work is already in progress, spread across multiple changes at the moment. This work sort of highlighted the problem. I can combine those PR's together but I would like to keep them split up so they're easier to review.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: Hadoop maintains a separate classpath that needs to be configured for each job in the driver program. We use the local job runners in the integration tests, but they still use their own classloaders when running the local job.

Historically, we configured the job jar (which contains the code for the integ test jobs, as well as the project code being tested) once on the test config. I'd like to move to using the original jar as-is without uber-jarring it with the itest code, but that will require updating every test case since it needs to be configured on every test job (can't be set on the configuration once and be done).

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: I'll remove this log line, it's not really needed or helpful

mr/src/itest/java/org/elasticsearch/hadoop/Provisioner.java
jbaiera: I totally agree, I'll fix this in a follow up.

jbaiera:
Relates #1423

PR #1441, opened 2020-03-13
Separate test utilities into a test library project

jbaiera:
We have a fair amount of testing mocks and utilities that are currently living in the MR project's itest source root. These mocks and utilities are used all over the project's integration test suites. For those projects to work, they currently break the glass and depend on the MR itest source root output directly. To fix this, there is now a shared test library subproject and all shared test code and test data has been moved to it.

This PR includes:
- Moved shared test source files to the `:test:shared` project.
- Add dependencies on that shared project to all projects' test configurations.
- Added a new TestData resource to unpack test input data into a temp directory (Hadoop does not like reading files from Jar URLs).
- Split the BuildPlugin (new plugin: BaseBuildPlugin) to separate global settings logic from the rest of the basic Java build configuration.


test/shared/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    implementation(project(':elasticsearch-hadoop-mr'))
+    
+    implementation(project.ext.hadoopClient)
```
mark-vieira: I suspect some of these probably _should_ be `api` dependencies as these test classes and fixtures expose Hadoop types in their APIs, [for example](https://github.com/jbaiera/elasticsearch-hadoop/blob/add-test-libs-project/test/shared/src/main/java/org/elasticsearch/hadoop/mr/PrintStreamOutputFormat.java#L38). This probably only compiles and works as-is incidentally because the consuming projects bring in hadoop already.

jbaiera: That's a good point. I'll put that in for correctness.

jbaiera:
Relates #1423

PR #1442, opened 2020-03-17
Fix project artifact sharing.

jbaiera:
Use a configuration to embed the contents of the `mr` project into dependent
projects' artifacts instead of using the `mr` project's jar task directly.

Consolidate all projects' dependency blocks into one block for each project.

### DIFF
--- a/hive/build.gradle
+++ b/hive/build.gradle
@@ -5,16 +5,38 @@ description = "Elasticsearch Hadoop Hive"
 
 evaluationDependsOn(':elasticsearch-hadoop-mr')
 
+configurations {
+    embedded {
+        transitive = false
+        canBeResolved = true
+    }
+    implementation {
+        extendsFrom project.configurations.embedded
+    }
+}
+
 dependencies {
-    provided(project(":elasticsearch-hadoop-mr"))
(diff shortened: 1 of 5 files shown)


hive/build.gradle
```diff
@@ -5,16 +5,38 @@ description = "Elasticsearch Hadoop Hive"
 
 evaluationDependsOn(':elasticsearch-hadoop-mr')
 
+configurations {
+    embedded {
+        transitive = false
+        canBeResolved = true
+    }
+    implementation {
+        extendsFrom project.configurations.embedded
```
jakelandis: I don't think we should change the `canBeResolved` for the implementation configuration. See https://discuss.gradle.org/t/what-is-a-configuration-which-cant-be-directly-resolved/30721/2 
> In short, you should never change the canBeConsumed and canBeResolved flags of configuration you do not create.

also, it's not clear you are changing it ?

jbaiera: Implementation shouldn't be affected by that at all. Only the `embedded` configuration is getting `canBeResolved` set on it. The `implementation` configuration just depends on it

jakelandis: ok , i mis-understood how `extendsFrom` behaves

jbaiera:
Relates #1423

PR #1443, opened 2020-03-17
Pass itest jar path to tests through system property

jbaiera:
This PR explicitly passes the location of the integration test jars to the integration test task instead of relying on the correct files existing in a relative path from the test's working directory.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -428,6 +428,9 @@ class BuildPlugin implements Plugin<Project>  {
 
             Test integrationTest = project.tasks.create('integrationTest', RestTestRunnerTask.class)
             integrationTest.dependsOn(itestJar)
+            integrationTest.doFirst {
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -428,6 +428,9 @@ class BuildPlugin implements Plugin<Project>  {
 
             Test integrationTest = project.tasks.create('integrationTest', RestTestRunnerTask.class)
             integrationTest.dependsOn(itestJar)
+            integrationTest.doFirst {
+                integrationTest.systemProperty("es.hadoop.job.jar", itestJar.getArchiveFile().get().asFile.absolutePath)
+            }
```
jbaiera: I could not get this to use a closure - It looks like Gradle was trying to serialize it, I'm assuming in order to cache the input values, but the serialization did not resolve the value into a string first.

mark-vieira: That is correct. Gradle essentially just calls `toString()` on these. One thing we've done in our build is just to create an anonymous object whose `toString()` implementation returns the value we want.

jbaiera:
Relates #1423

PR #1444, opened 2020-03-18
Fix BlockAwareJsonParser level accounting

jbaiera:
The BlockAwareJsonParser is used to set a checkpoint on a JSON stream
and exit the current blocks that the cursor is in back to the same
level of the content that it was created at. If skipChildren is called
to skip over an open array or object, the internal level counter will
not be updated to account for the array or object that was skipped over.

This PR updates the skipChildren method to check the currentToken and
conditionally decrement the internal nested level counter if we ended
up at an end of an array or object.

### DIFF
--- a/mr/src/test/java/org/elasticsearch/hadoop/serialization/json/BlockAwareJsonParserTest.java
+++ b/mr/src/test/java/org/elasticsearch/hadoop/serialization/json/BlockAwareJsonParserTest.java
@@ -176,4 +176,34 @@ public class BlockAwareJsonParserTest {
         blockParser.exitBlock();
         assertThat(parser.currentToken(), nullValue());
     }
+
+    /**
+     * We increment the level of nesting in the parser when getting the next token. If that token starts an array or object,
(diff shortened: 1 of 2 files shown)


mr/src/test/java/org/elasticsearch/hadoop/serialization/json/BlockAwareJsonParserTest.java
```diff
@@ -176,4 +176,34 @@ public void testSkippingAndEncounterEOF() {
...
+    /**
+     * We increment the level of nesting in the parser when getting the next token. If that token starts an array or object,
+     * the "open" counter is incremented. If we then call `skipChildren` instead of iterating to the end of the object, make
+     * sure that the open counter is decremented.
+     */
+    @Test
+    public void testExitBlockAfterSkippingChildren() {
+        String data = "{\"nested\":{\"array\":[\"test\"],\"scalar\":1}}";
+        //                         ^          !        |-------------^
+        // ! = skipChildren
```
dakrone: :heart:

Issue #1456, opened 2020-04-02
[CI] Test failure on 7.x branch: AbstractScalaEsScalaSparkSQL testDataSourcePushDown12And


williamrandolph:
This test has been failing on the 7.x branch since the version bump in Elasticsearch ( [jenkins](https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch-hadoop+7.x+periodic/) ).

I was not able to reproduce locally because I ran into different build errors.

## Sample stacktrace

Here's one of the stack traces attached to the email.

```
java.lang.AssertionError: expected:<0> but was:<2>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.elasticsearch.spark.integration.AbstractScalaEsScalaSparkSQL.testDataSourcePushDown09StartsWith(AbstractScalaEsSparkSQL.scala:1001)
	at sun.reflect.GeneratedMethodAccessor106.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
[...]

PR #1458, opened 2020-04-03
Update Spark tests that use wildcard query

jbaiera:
There were some issues fixed with the Wildcard query in https://github.com/elastic/elasticsearch/pull/53127 that now cause a few tests to fail in ES-Hadoop. Looking through the logic, it seems that the tests were originally written with the incorrect behavior in mind. This PR updates them to run again.


jbaiera:
Fixes #1456

PR #1460, opened 2020-04-09
Upgrade to Gradle 6.3 and Java 14

jbaiera:
Elasticsearch is now running its builds on Java 14. This PR upgrades to Gradle 6.3 and Java 14 to keep in sync.


jbaiera:
Relates #1423

PR #1461, opened 2020-04-09
Import sources for javadoc through a dependency

jbaiera:
We embed project code from the MR project into each and every integration Project in ES-Hadoop. When generating javadocs, we directly reference the mr project's source sets to pull their contents into the javadoc tool as an input. 

This becomes a problem when a downstream project depends on java source code that is generated by a task. A specific instance of this problem arises when splitting the Spark core and SQL projects apart. The Spark projects take advantage of a compiler plugin to generate java sources specifically to be used with the Javadoc tool. These generated sources must be transmitted from the Spark core projects to the Spark SQL projects so that they may be bundled together in the same documentation package.

This PR adds configurations for importing and exporting source directories. It adds a `sourceElements` configuration that can be consumed by other projects. The build plugin then registers the main java source directories in the project with this configuration through the artifacts handler. The PR also adds a `sources` configuration that can be resolved. Projects use this configuration to denote source code directories from other projects that are needed to generate the Javadocs and Sources jar.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -91,7 +91,7 @@ class BuildPlugin implements Plugin<Project>  {
         if (project != project.rootProject) {
             // Set up avenues for sharing source files between projects in order to create embedded Javadocs
             // Import source configuration
-            Configuration sources = project.configurations.create("sources")
+            Configuration sources = project.configurations.create("additionalSources")
             sources.canBeConsumed = false
             sources.canBeResolved = true
             sources.attributes {
@@ -276,7 +276,7 @@ class BuildPlugin implements Plugin<Project>  {
         sourcesJar.from(project.sourceSets.main.allSource)
         // TODO: Remove when root project does not handle distribution
         if (project != project.rootProject) {
-            sourcesJar.from(project.configurations.sources)
+            sourcesJar.from(project.configurations.additionalSources)
         }
 
         // Configure javadoc
@@ -291,7 +291,7 @@ class BuildPlugin implements Plugin<Project>  {
         ]
         // TODO: Remove when root project does not handle distribution
         if (project != project.rootProject) {
-            javadoc.source += project.configurations.sources
+            javadoc.source = project.files(project.configurations.additionalSources)
         }
         // Set javadoc executable to runtime Java (1.8)
         javadoc.executable = new File(project.ext.runtimeJavaHome, 'bin/javadoc')
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -17,11 +17,15 @@ import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.ResolutionStrategy
 import org.gradle.api.artifacts.maven.MavenPom
 import org.gradle.api.artifacts.maven.MavenResolver
+import org.gradle.api.attributes.LibraryElements
+import org.gradle.api.attributes.Usage
 import org.gradle.api.file.CopySpec
+import org.gradle.api.file.FileCollection
 import org.gradle.api.java.archives.Manifest
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.plugins.MavenPlugin
 import org.gradle.api.plugins.MavenPluginConvention
+import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.Upload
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -200,6 +235,15 @@ class BuildPlugin implements Plugin<Project>  {
         project.sourceCompatibility = '1.8'
         project.targetCompatibility = '1.8'
 
+        // TODO: Remove all root project distribution logic. It should exist in a separate dist project.
```
jbaiera, line 238: This is a problem that I'm working on addressing in another PR. The root project is all sorts of messed up because it depends (without using dependencies in a lot of cases 😱) on all the subprojects to build an uber jar. This uberjar/dist zip should be generated in a subproject instead of the root project, using dependencies, configurations, and all other appropriate build script hygiene to resolve the cross project links.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -84,6 +88,32 @@ class BuildPlugin implements Plugin<Project>  {
     }
 
     private static void configureConfigurations(Project project) {
+        if (project != project.rootProject) {
+            // Set up avenues for sharing source files between projects in order to create embedded Javadocs
+            // Import source configuration
+            Configuration sources = project.configurations.create("sources")
+            sources.canBeConsumed = false
+            sources.canBeResolved = true
```
jakelandis, line 96: this is be false too ?  the sources shouldn't be resolved since they should only be declared as a single project reference ?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -84,6 +88,32 @@ class BuildPlugin implements Plugin<Project>  {
     }
 
     private static void configureConfigurations(Project project) {
+        if (project != project.rootProject) {
+            // Set up avenues for sharing source files between projects in order to create embedded Javadocs
+            // Import source configuration
+            Configuration sources = project.configurations.create("sources")
```
jakelandis, line 94: would `additionalSources` read better ?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -200,6 +230,15 @@ class BuildPlugin implements Plugin<Project>  {
         project.sourceCompatibility = '1.8'
         project.targetCompatibility = '1.8'
 
+        // TODO: Remove all root project distribution logic. It should exist in a separate dist project.
+        if (project != project.rootProject) {
+            SourceSet mainSourceSet = project.sourceSets.main
+            FileCollection javaSourceDirs = mainSourceSet.java.sourceDirectories
+            javaSourceDirs.each { File srcDir ->
```
jakelandis, line 237: is this loop correct ?  I would have expected to see something like:
```
project.getArtifacts().add('sourceElements', mainSourceSet.java.sourceDirectories)
```
(no loop)

spark/sql-20/build.gradle
```diff
@@ -156,6 +156,15 @@ dependencies {
     itestImplementation("org.apache.spark:spark-streaming_${project.ext.scalaMajorVersion}:$sparkVersion") {
         exclude group: 'org.apache.hadoop'
     }
+
+    sources(project(":elasticsearch-hadoop-mr"))
+}
+
+// Export generated Java code from the genjavadoc compiler plugin
+artifacts {
+    sourceElements(project.file("$buildDir/generated/java")) {
```
jakelandis, line 165: does adding the `sourceElements` here conflict with plugin addition ?  ( I don't know if it is additive or a replacement)

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: Yeah I think you're on to something there. I'll make that change.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
mark-vieira: I don't believe that will work since `sourceDirectories` is a `FileCollection`. The `ArtifactHandler.add()` doesn't accept collections. You need to add the directories individually.

That said, because of this, we are open to configuration ordering issues here. So any additional source directories need to be added _before_ this code runs.

spark/sql-20/build.gradle
mark-vieira: This isn't _adding_ a configuration. It's adding an artifact to the existing `sourceElements` configuration.

jbaiera: Not sure what you mean by this question. Do you mean the scala compiler plugin or the build plugin?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: Hmm, gradle does not seem to like this. In order for Gradle to convert the project dependency into actual files it needs to resolve the configuration. It needs to do this for the sources jar, and any time a file collection that contains the configuration is mutated.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: Yeah, I wasn't able to find anything on FileCollections that allows you to listen for changes like other domain object collections. I think anything outside of the default java directory needs to be added to the source set explicitly, and thus should be getting added to the sourceElements configuration explicitly as well.

jbaiera:
Relates #1423

PR #1463, opened 2020-04-14
Use subproject to generate distribution instead of root

jbaiera:
This PR adds a subproject (`dist`) which collects the contents of all integrations together into the project uberjar, and the zip distribution. Currently, these are created in the root project, mostly without proper dependency modeling. 

Creating as Draft to collect some feedback.


dist/build.gradle
```diff
@@ -0,0 +1,141 @@
...
+    // To squash Javadoc warnings
+    compileOnly(project(":elasticsearch-hadoop-mr"))
+    compileOnly(project(":elasticsearch-hadoop-hive"))
+    compileOnly(project(":elasticsearch-hadoop-pig"))
+    compileOnly(project(":elasticsearch-spark-20"))
+    compileOnly(project(":elasticsearch-storm"))
+    compileOnly("org.apache.spark:spark-catalyst_${project.ext.scala211MajorVersion}:$spark20Version")
+
+    // For Uber pom (and Javadoc to a lesser extent)
+    // TODO: Would be way better if we could pull these directly from those projects. Should we just grab the configurations?
```
jbaiera, line 39: I'm pulling each dependency explicitly in order to get the POM file to generate the correct dependencies. Using regular project dependencies just makes the POM file reference the artifacts from those projects instead of the dependencies from those projects.

dist/build.gradle
```diff
@@ -0,0 +1,141 @@
...
+    additionalSources(project(":elasticsearch-spark-20"))
+    additionalSources(project(":elasticsearch-storm"))
+
+    // This will pull in the regular jar, javadoc jar, and source jar for each project
+    dist(project(":elasticsearch-hadoop-mr"))
+    dist(project(":elasticsearch-hadoop-hive"))
+    dist(project(":elasticsearch-hadoop-pig"))
+    dist(project(":elasticsearch-spark-20"))
+    dist(project(":elasticsearch-storm"))
+}
```
jbaiera, line 93: We have a lot of repeated project dependencies here. Each configuration is pulling in something different:

`embedded` is pulling in just the java jar file from each project without transitive dependencies in order to create the uber jar.
`additionalSources` is pulling in sources meant for javadoc generation
`dist` is pulling in each integration's jar, javadocJar, and sourceJar to be packed up in the zip file for the download page.

We might need to add a different sources configuration to pull in original sources (scala and java) instead of the current state (java and generated java) for the sake of having an accurate sourcesJar.

This seems like a ton of repeated project dependencies. Is this the right way to do things?

dist/build.gradle
mark-vieira: If you want to bring in the dependencies too, then teh configuration you consume needs to extend the correct configuration. The `default` configuration already does this by extending `runtime`.

dist/build.gradle
```diff
@@ -0,0 +1,141 @@
...
+    embedded {
+        canBeResolved = true
+        canBeConsumed = false
+        transitive = false
+    }
+    dist {
+        canBeResolved = true
+        canBeConsumed = false
+        attributes {
+            attribute(Usage.USAGE_ATTRIBUTE, objects.named(Usage, 'packaging'))
```
mark-vieira, line 17: Brilliant.

dist/build.gradle
mark-vieira: You could define the set of projects once and then just loop over them adding them to each different configuration.

dist/build.gradle
jbaiera: After thinking about this for a little bit, having these duplicated here isn't too bad, especially since many of the dependencies might change scopes on the uberpom compared to each integration's pom file.

dist/build.gradle
jbaiera: This did end up being better in the long run. I ended up needing to define a couple more configurations to get javadocs and source jars generating correctly due to the mixed java and scala sources.

jbaiera:
I think I'm good for review on this now. I left in the current root project distribution logic for now since release manager depends on it. Once this is merged, I'll make the changes to release manager to look at the new location for distribution artifacts, then follow up on this project to remove the old root project distribution tasks.

dist/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+    options {
+        header = "elasticsearch-hadoop"
+    }
+}
+
+// Name of the directory under the root of the zip file that will contain the zip contents
+String zipContentDir = "elasticsearch-hadoop-${project.version}"
+
+// Create a zip task for creating the distribution
+task('distZip', type: Zip.class) {
```
mark-vieira, line 107: nit: this can just be `type: Zip`. No need for the `.class`.

jbaiera:
Relates #1423

PR #1471, opened 2020-05-12
Upgrade to Gradle 6.4

jbaiera:
This version of gradle includes the fix for Scala's fork compile options that will allow us to control which version of the JVM will run the Scala compiler, re-enabling the ability to compile for Scala 2.10, and opening the doors to backporting our build changes to the 7.x branch.

This PR includes the gradle upgrade, re-enabling the scala 2.10 compilation (for ease of backporting our lineup of changes), as well as a small fix to correct a compilation error in the QA Kerberos project due to how the legacy variant compilation process works.


jbaiera:
Relates #1423

PR #1481, opened 2020-05-28
Fix Scala test compile Java home usage

jbaiera:
Setting the java home property was only done on the regular Scala compile task. It should be set on all Scala compile tasks instead.

### DIFF
--- a/spark/sql-13/build.gradle
+++ b/spark/sql-13/build.gradle
@@ -55,10 +55,10 @@ configurations.all { Configuration conf ->
 tasks.withType(ScalaCompile) { ScalaCompile task ->
     task.sourceCompatibility = project.ext.minimumRuntimeVersion
     task.targetCompatibility = project.ext.minimumRuntimeVersion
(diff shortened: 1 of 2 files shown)


spark/sql-13/build.gradle
```diff
@@ -55,10 +55,10 @@ configurations.all { Configuration conf ->
 tasks.withType(ScalaCompile) { ScalaCompile task ->
     task.sourceCompatibility = project.ext.minimumRuntimeVersion
     task.targetCompatibility = project.ext.minimumRuntimeVersion
+    task.options.forkOptions.executable = new File(project.ext.runtimeJavaHome, 'bin/java').canonicalPath
```
mark-vieira, line 58: Is there a compelling reason to use `canonicalPath` here? There's a performance cost to doing so as resolving a canonical path involves file IO at configuration time.

jbaiera: I'll put up a quick fix for this then. I was following the same approach that the Gradle internals are doing, but I suppose if they're already getting the canonical file, we don't have to here.

jbaiera:
Relates #1423

PR #1508, opened 2020-08-05
Update references to renamed Gradle task

mark-vieira:
Some of our rest testing tasks were renamed as part of the refactoring done in https://github.com/elastic/elasticsearch/pull/60261. This broke the hadoop build as it still contains references to those old tasks. Right now this is only failing in our release builds due to the new `build-tools` jar not being published yet but we need to get those builds green to get the updated artifact published.


mark-vieira:
Yes, I was able to run `./gradlew integrationTest` locally.

mark-vieira:
@jbaiera We good to merge on this?

mark-vieira:
I'm going to hit the button on this to confirm this fixes the release snapshot builds.

mark-vieira:
This has also been backported to `7.x`.

Issue #1527, opened 2020-09-18
array type default is array and has one null value


ZhendongBai:
### What kind an issue is this?

- when insert null value, array type default is array and has one null value, and how to handle the unset value

### Issue description

- **why the unset value(i.e. the first row of result) status_code has the array(null) as its default value ?**
- **how to handle the default unset value ?**

### Steps to reproduce

Code:

```
PUT my-index-000001/_doc/1
{
  "status_code": null,
  "ts": 12345678910,
  "name": "john"
}

PUT my-index-000001/_doc/2
{
  "status_code": ['123'],
  "ts": null,
  "name": "johnny"
}

PUT my-index-000001/_doc/3
{
  "status_code": [],
  "ts": 12345678910,
  "name": null
}
```
the result show as following:
![image](https://user-images.githubusercontent.com/18043146/93551135-b92be780-f99f-11ea-9322-f6cbb1813d84.png)

query the collection result with spark sql:
```
queryStatement =
      """SELECT status_code,
        | size(status_code),
        | case when size(status_code) = 1 then status_code[0] else null end as first_status_code,
        | ts,
        | name
        | FROM artists""".stripMargin
    df = spark.sql(queryStatement)
    df.show(200, truncate = false)
``` 
the result shows as below:
```
+-----------+-----------------+-----------------+-----------+------+
|status_code|size(status_code)|first_status_code|ts         |name  |
+-----------+-----------------+-----------------+-----------+------+
|[]         |1                |null             |12345678910|john  |
|[123]      |1                |123              |null       |johnny|
|[]         |0                |null             |12345678910|null  |
+-----------+-----------------+-----------------+-----------+------+
```
### Version Info

OS:         :  mac pro
JVM         :  1.8
Hadoop/Spark:  2.4.3
ES-Hadoop   :  7.9.1
ES          :  7.5.2(docker container single-node)

jbaiera:
The existence of a null field value can be interpreted in both ways equally validly, but the absence of the field is evidence enough to indicate that the array itself should be null. Classifying as a bug.

johnnyprimary:
@jbaiera  actually, I have signed the CLA, but the CLA service tells me I am not sign the CLA....

masseyke:
I get different results following your steps. @ZhendongBai can you tell me what I'm doing differently? I indexed 3 documents into the my-index-000001 table (which did not previously exist) as you show. Then I did the following in spark shell:
```
[...]

PR #1531, opened 2020-09-21
bugfix: when array value is not set, the default value contains one element - null value.

johnnyprimary:
bugfix ： https://github.com/elastic/elasticsearch-hadoop/issues/1527

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/serialization/ScrollReader.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/serialization/ScrollReader.java
@@ -981,7 +981,9 @@ public class ScrollReader implements Closeable {
         Object array = reader.createArray(mapping(fieldMapping, parser));
         // create only one element since with fields, we always get arrays which create unneeded allocations
         List<Object> content = new ArrayList<Object>(1);
-        content.add(value);
+        if (null != value) {
+            content.add(value);
+        }
         array = reader.addToArray(array, content);
         return array;
     }


johnnyprimary:
> ❌ Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [78b2cfb](https://github.com/elastic/elasticsearch-hadoop/commit/78b2cfbdda3d89be7030af917b66b7ee6ff4842d)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

actually， I have signed the CLA

jbaiera:
Please make sure that the email you signed the CLA with matches your github account as well as the email the commits were authored under. I'll see if the CLA maintainers can weigh in on if there is a problem.

Additionally, thanks for opening a PR for this. Can you add an additional test case for this bug?

mr/src/main/java/org/elasticsearch/hadoop/serialization/ScrollReader.java
```diff
@@ -981,7 +981,9 @@ protected Object singletonList(String fieldMapping, Object value, Parser parser)
         Object array = reader.createArray(mapping(fieldMapping, parser));
         // create only one element since with fields, we always get arrays which create unneeded allocations
         List<Object> content = new ArrayList<Object>(1);
-        content.add(value);
+        if (null != value) {
+            content.add(value);
```
jbaiera, line 985: This is still going to create an empty array for a field that is listed as null. Is this the outcome you'd always expect to see instead of just setting the field to null?

Personally, I can see an array field set to null as equally meaning no array at all. I think we need to hammer down what the correct default should be. Either an empty array or a null field, but I agree that it should not be an array with a null value.

jbaiera:
Following up on the CLA, it looks like this PR is from a different author account on Github than the one that has opened the PR. Are you the original author of this patch? You may need to open the PR with the account that the commit author is registered as.

johnnyprimary:
> Please make sure that the email you signed the CLA with matches your github account as well as the email the commits were authored under. I'll see if the CLA maintainers can weigh in on if there is a problem.
> 
> Additionally, thanks for opening a PR for this. Can you add an additional test case for this bug?

ok, I will add a test case for this bug.

jbaiera:
Closing this as there has been no further progress.

PR #1544, opened 2020-10-20
Update Gradle wrapper to 6.7

breskeby:
- I just updated the elasticsearch build to use gradle 6.7 so we should align this es-hadoop build too.
See https://github.com/elastic/elasticsearch/pull/62386


jbaiera (approved):
LGTM, thanks!

PR #1557, opened 2020-11-20
Update references to internal precommit plugins

mark-vieira:
This pull request updates some references to refactored tasks in build-tools. Ironically, the cause here is that we are undergoing an effort in Elasticsearch to make better separation between "internal" and "external" build logic, specifically to minimize fallout to external users, such as es-hadoop. The root cause PR is https://github.com/elastic/elasticsearch/pull/65102.

@jbaiera essentially the new contract is that if you use stuff in an `internal` package we make no breaking changes guarantees. The preferred approach here would be to remove to dependencies on anything `internal` in build-tools. In fact, we eventually intend to remove that stuff entirely and simply not include them in the published build-tools jar.


jbaiera (approved):
LGTM

mark-vieira:
Backported to `7.x`.

PR #1584, opened 2021-01-19
Add properties for pulling java versions from the environment variables

jbaiera:
CI is failing at the moment because it cannot locate java installations with the GlobalBuildInfoPlugin.


mark-vieira (approved):
:+1:

PR #1586, opened 2021-01-20
Distinguish core Spark artifacts from each other

jbaiera:
The new tooling for building Spark/Scala variants seems to have an issue with the core Spark project. This project contains multiple versions of Scala _and_ Spark. The artifacts produced from here are differentiated in name only by the version of Scala. This means that building for Spark 2.0 on 2.11 and Spark 1.6 on 2.11 would produce the same artifact name. Luckily, the core code does not change between these versions and no compatibility issues have yet been raised because of this bug.

This PR updates the Spark integration's core project to prepend the Spark version of its variant to the artifacts classifier as it normally would (we strip this out in the plugin code since the classifier is unneeded for the final products).


danhermann (approved):
LGTM. Thanks, @jbaiera!

PR #1600, opened 2021-02-16
Fix dependency licenses for Spark 3

jbaiera:
Update the dependency license versions and SHAs for the Spark 3 upgrade.


jakelandis (approved):
LGTM (however, i did not recalculate these to confirm)

Issue #744, opened 2016-04-13
Upgrade to Hive 2.0


costin:
Title says it all.

costin:
Unfortunately it looks like Hive 2.0 dropped compatibility with Hadoop 1.x (unclear why). As most code bases are still on Hive 1.x, the upgrade will be postponed.
The code is compatible but by dropping support for Hadoop 1.x it only complicates the building process (since Hive would have to compiled only when running against Hadoop 2.x).

rickw1028:
Hi guys, Do you have any ETA to support Hive 3.X ?

Issue #1300, opened 2019-06-07
[feature request]: add support for Hadoop 3.0+


schories:
Feature Request:

As CDH 6.x.x comes with Hadoop 3.0+ 

https://www.cloudera.com/documentation/enterprise/6/release-notes/topics/rg_cdh_62_packaging.html#cdh_packaging_60x

official support for Hadoop 3.0+ is now required.

ankurthuse:
Is there any update on this? Have a meeting with customer next week and they are asking if there is any update on this.

Issue #1004, opened 2017-06-16
Upgrade to Hadoop 2.8


jbaiera:
Hadoop 2.8.0 was released on March 22 of this year, with a 2.8.1 release in process (at the time this ticket was created). Most notably in the list of changes for Hadoop 2.8.0 is the complete replacement of the apache commons httpclient 3.1 with the newer org.apache.httpcomponents httpclient 4.5.2. This poses a problem for upgrading since in 6.0 we will be supporting all the way back to Hadoop 2.2.0.

zackehh:
Just dropping a specific error I'm seeing for reference in case people are searching for it. Can confirm that it's caused by 2.8.x.

```
java.lang.Exception: java.lang.NoClassDefFoundError: org/apache/commons/httpclient/HttpConnectionManager
    at org.apache.hadoop.mapred.LocalJobRunner$Job.runTasks(LocalJobRunner.java:489)
    at org.apache.hadoop.mapred.LocalJobRunner$Job.run(LocalJobRunner.java:556)
Caused by: java.lang.NoClassDefFoundError: org/apache/commons/httpclient/HttpConnectionManager
    at org.elasticsearch.hadoop.rest.commonshttp.CommonsHttpTransportFactory.create(CommonsHttpTransportFactory.java:39)
    at org.elasticsearch.hadoop.rest.NetworkClient.selectNextNode(NetworkClient.java:99)
    at org.elasticsearch.hadoop.rest.NetworkClient.<init>(NetworkClient.java:82)
    at org.elasticsearch.hadoop.rest.NetworkClient.<init>(NetworkClient.java:59)
    at org.elasticsearch.hadoop.rest.RestClient.<init>(RestClient.java:92)
    at org.elasticsearch.hadoop.rest.InitializationUtils.discoverEsVersion(InitializationUtils.java:273)
    at org.elasticsearch.hadoop.rest.RestService.createWriter(RestService.java:572)
    at org.elasticsearch.hadoop.mr.EsOutputFormat$EsRecordWriter.init(EsOutputFormat.java:173)
    at org.elasticsearch.hadoop.mr.EsOutputFormat$EsRecordWriter.write(EsOutputFormat.java:149)
    at org.apache.hadoop.mapred.ReduceTask$NewTrackingRecordWriter.write(ReduceTask.java:558)
    at org.apache.hadoop.mapreduce.task.TaskInputOutputContextImpl.write(TaskInputOutputContextImpl.java:89)
[...]

PR #1636, opened 2021-03-30
Upgrade Hadoop and Hive to version 3

jbaiera:
This PR upgrades our dependencies on Hadoop and Hive to version 3.x.

Hadoop removed and isolated some dependencies that we make use of in the connector, primarily the HTTP client libraries. In order to maintain a single-jar deployment model, we are now shading the client library dependencies into our release artifacts. Notice files and licenses checking has been updated for this.

Our version of Hive was too old to function on the newer version of Hadoop, so Hive has also been updated to version 3.x on its own release track. I have tested that the hive integration code still works with Hive 1.2, but in the next major version we may be updating the minimum supported version of Hive.

The Hadoop version upgrade also caused some issues with our Spark testing code. Hadoop has upgraded the version of Jackson it depends on. The version that Hadoop provides conflicts with the version of Jackson that Spark expects. This is normally not a problem when deploying either solutions in isolation, but causes compatibility issues with Jackson modules when attempting to support both frameworks with one library. As such, the Jackson libraries required have also been shaded into the release artifacts for ES-Hadoop. We already packaged some Jackson code for backwards compatibility, so NOTICE files are already valid for this change.

A number of changes have gone in to support the version updates in the QA project for Kerberos. A number of testing changes were required in the integration tests for Hive due to changes to how the Hive metastore is initialized. Some additional changes to support running on external clusters also went in.


pig/src/itest/java/org/elasticsearch/hadoop/integration/pig/AbstractPigExtraTests.java
```diff
@@ -104,13 +145,13 @@ public void testIterate() throws Exception {
         RestUtils.refresh("pig-test-iterate");
 
         String script =
-                "REGISTER "+ Provisioner.ESHADOOP_TESTING_JAR + ";" +
+//                "REGISTER "+ Provisioner.ESHADOOP_TESTING_JAR + ";" +
```
jbaiera, line 148: If we don't actually need these any more, they should probably be thrown out.

pig/src/itest/resources/log4j2.properties
```diff
@@ -0,0 +1,80 @@
...
+logger.Hive-ExecMapper.name=ExecMapper
+logger.Hive-ExecMapper.level=warn
+logger.Hive-HiveQL-Package.name=org.apache.hadoop.hive.ql
+logger.Hive-HiveQL-Package.level=warn
+logger.Hive-HiveQL-Driver.name=org.apache.hadoop.hive.ql.Driver
+logger.Hive-HiveQL-Driver.level=warn
+logger.Hive-RetryingHMSHandler.name=org.apache.hadoop.hive.metastore.RetryingHMSHandler
+logger.Hive-RetryingHMSHandler.level=fatal
+logger.Hive-HiveQL-Exec-Package.name=org.apache.hadoop.hive.ql.exec
+logger.Hive-HiveQL-Exec-Package.level=warn
```
jbaiera, line 25: Maybe ditch the hive loggers for Pig

buildSrc/build.gradle
```diff
@@ -77,6 +77,7 @@ dependencies {
     } else {
         compile group: 'org.elasticsearch.gradle', name: 'build-tools', version: buildToolsVersion
     }
+    compile group: 'com.github.jengelman.gradle.plugins', name: 'shadow', version: '6.1.0'
```
jbaiera, line 80: Is this needed if we're just using it in the thirdparty project now?

jbaiera (commented):
Combed through the changes and found a few last clean up spots

jbaiera:
fixes #744 #1004 #1300

pan3793:
Consider switch to Hadoop Shaded Client? https://issues.apache.org/jira/browse/SPARK-33212

jbaiera (commented):
Feedback during second review

jbaiera:
> Consider switch to Hadoop Shaded Client? https://issues.apache.org/jira/browse/SPARK-33212

This might be a good thing for us to depend on in the project instead of the regular client libraries we pull in, but I think it should be a part of another PR - this one is already too large.

I think that shading our dependencies ourselves going forward will give us more flexibility to upgrade them without the concern that they'll break other integrations or user code, but the API dependency could be a good way to avoid pulling in any conflicting transitive dependencies.

pan3793:
> I think that shading our dependencies ourselves going forward will give us more flexibility to upgrade them without the concern that they'll break other integrations or user code, but the API dependency could be a good way to avoid pulling in any conflicting transitive dependencies.

The official shaded client only expose public Hadoop API which quite stable. And the shade behavior is not always safe, maybe accident broken some libraries which depends the specific Hadoop API, even the official shaded client can not do it perfect, i.e. https://github.com/apache/hadoop/pull/2575

For my perspective, I recommend to make the project only depends on Hadoop public API, and test against Hadoop Shaded Client(>=3.0) and normal Hadoop Client(< 3.0), which just Apache Spark doing now.

jbaiera:
> And the shade behavior is not always safe, maybe accident broken some libraries which depends on the specific Hadoop API

We aren't shading any artifacts from the Hadoop packages. We're shading dependencies that Hadoop previously provided to jobs at runtime. If we were shading Hadoop dependencies, I for sure would agree with you - Any change to those API's in the runtime can cause runtime problems, but as it stands only HTTP Client and Jackson are being shaded in ES-Hadoop.

jakelandis (approved):
LGTM

(Jimmy walked Lee and myself through the changes and all tests pass locally)

jakelandis:
I noticed a minor issue with the build with local repo vs. not local.  Can you add the following (verify by running ./gradlew clean && ./gradlew clean -PlocalRepo)

```diff
diff --git a/settings.gradle b/settings.gradle
index 259e6e31..e47115ed 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,9 @@
+pluginManagement {
+    plugins {
+        id 'com.github.johnrengelman.shadow' version "6.1.0"
+    }
+}
+
 rootProject.name = "elasticsearch-hadoop"
 
 include 'thirdparty'
@@ -37,3 +43,4 @@ include 'test:fixtures:minikdc'
 
 include 'qa'
 include 'qa:kerberos'
+
diff --git a/thirdparty/build.gradle b/thirdparty/build.gradle
index 03e491b2..c0717fd0 100644
--- a/thirdparty/build.gradle
+++ b/thirdparty/build.gradle
@@ -1,7 +1,7 @@
 import org.elasticsearch.hadoop.gradle.BuildPlugin
 
 plugins {
-    id 'com.github.johnrengelman.shadow' version "6.1.0"
+    id 'com.github.johnrengelman.shadow'
     id 'es.hadoop.build'
 }
```

jbaiera:
I've updated the plugin management so that it works with and without local repo now.

PR #1644, opened 2021-04-15
Fix the build's usage of the node roles settings

jbaiera:
This PR changes the test fixture plugin to use the new `node.roles` setting instead of `node.ingest`


jakelandis (approved):
LGTM

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/ElasticsearchFixturePlugin.groovy
```diff
@@ -72,9 +72,13 @@ class ElasticsearchFixturePlugin implements Plugin<Project> {
             integTestCluster.setting("node.ingest", "true")
             integTestCluster.setting("http.host", "localhost")
             integTestCluster.systemProperty('es.http.cname_in_publish_address', 'true')
-        } else if (majorVersion >= 7) {
+        } else if (majorVersion == 7) {
```
dakrone, line 75: Why not do it for 7.x also, since then at least we're not using deprecated functionality?

PR #1654, opened 2021-04-23
Do not rely on internal build tool logic

breskeby:
This PR detangles the hadoop build from the internal classes of elasticsearch build tools.
Mostly it copies over (simplified) plugins and tasks used by hadoop that are not supported
by build tools and not be puslished in the future to be used by external plugins.
In the longterm we probably will regrow the public surface of the elasticsearch build tools but
for now we basically only want to support TestCluster usage and a basic version of the
PluginBuildPlugin to support external projects building elasticsearch plugins.

This is related to https://github.com/elastic/elasticsearch/issues/71593

Tested manually here: //gradle-enterprise.elastic.co/s/6xbuknj4e24u2

- [x ] I have signed the [Contributor License Agreement (CLA)][]

### DIFF
--- a/buildSrc/src/main/java/org/elasticsearch/hadoop/gradle/buildtools/info/BuildParams.java
+++ b/buildSrc/src/main/java/org/elasticsearch/hadoop/gradle/buildtools/info/BuildParams.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+package org.elasticsearch.hadoop.gradle.buildtools.info;
+
+import org.gradle.api.JavaVersion;
+
+import java.io.File;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+public class BuildParams {
+    private static File runtimeJavaHome;
+    private static Boolean isRuntimeJavaHomeSet;
+    private static List<JavaHome> javaVersions;
+    private static JavaVersion minimumCompilerVersion;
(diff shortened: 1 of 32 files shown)


mark-vieira (approved):
LGTM.

buildSrc/src/main/java/org/elasticsearch/hadoop/gradle/buildtools/info/BuildParams.java
```diff
@@ -0,0 +1,156 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
```
jbaiera, line 7: We may need to double check if these files are ok to be listed with the Elastic license or if they need to be converted to Apache for this project. It's likely that this license is fine since it does not get published from the project as an artifact, but it should be double checked for compliance.

jbaiera (commented):
Things look in order to me, but I noticed one thing about the licenses that we may need to follow up on:

breskeby:
good point. I double check with @Mark Vieira ***@***.***>

buildSrc/src/main/java/org/elasticsearch/hadoop/gradle/buildtools/info/BuildParams.java
mark-vieira: We can relicense this as Apache 2.0 for simplicity. We are the copyright owners so we are free to do so at our discretion.

breskeby:
Updated the licenses to apache 2 as the rest of the sources in this repo.

buildSrc/src/main/java/org/elasticsearch/hadoop/gradle/buildtools/info/BuildParams.java
breskeby: Updated the licenses to apache 2 as the rest of the sources in this repo.

breskeby:
cherry picked this into 7.x branch so I can merge the repackaging on elasticsearch also down to 7.x without breaking the Hadoop build

PR #1661, opened 2021-05-18
Fix test cluster wiring for integTests

breskeby:
We brought back the StandaloneRestTestTask as part of the public api of build-tools
and do not require a hadoop internal copy of that task type.
Replacing the internal copy by the public api task fixes the test cluster wiring
for integration tests in the hadoop build.


jbaiera (approved):
LGTM! Thanks for tracking this down!

williamrandolph:
I think we do need to backport this for Elasticsearch Hadoop 7.x periodic builds: https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch-hadoop+7.x+periodic/

breskeby:
back ported this to 7.x

PR #1679, opened 2021-06-01
Verify cluster main info response returns correct product headers

### LINKED ISSUE #1672: Validate server headers during startup
Eventually Elasticsearch will transmit a header on all responses that will confirm the type of service running on the server side of the connection. We should validate this header at client start up to avoid possible formatting/protocol/validation issues going forward. 

See also https://github.com/elastic/elasticsearch/issues/73424

jbaiera:
This PR ensures that ES-Hadoop is running against a verified Elasticsearch deployment during start up. The check ensures response format compatibility and protocol compliance with the server. The change also has ES-Hadoop warn on mismatched internal configurations for the cluster.

Closes #1672


martijnvg (approved):
LGTM 👍

PR #1683, opened 2021-06-09
Fix failing strict wildcard pushdown tests

jbaiera:
A change that went in a while ago has made changes to the wildcard query functionality. It now functions as it had historically. This PR updates the testing logic to make sure the wildcard tests account for these differences.

### DIFF
--- a/spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
+++ b/spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
@@ -1005,8 +1005,12 @@ class AbstractScalaEsScalaSparkSQL(prefix: String, readMetadata: jl.Boolean, pus
     }
 
     filter.show
-    assertEquals(1, filter.count())
-    assertEquals("feb", filter.select("tag").take(1)(0)(0))
+    if (strictPushDown) {
+      assertEquals(0, filter.count()) // Strict means specific terms matching, and the terms are lowercased
+    } else {
+      assertEquals(1, filter.count())
+      assertEquals("feb", filter.select("tag").take(1)(0)(0))
+    }
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


jbaiera:
See also https://github.com/elastic/elasticsearch/pull/71751

spark/sql-13/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
```diff
@@ -1036,7 +1044,7 @@ class AbstractScalaEsScalaSparkSQL(prefix: String, readMetadata: jl.Boolean, pus
   @Test
   def testDataSourcePushDown12And() {
     val df = esDataSource("pd_and")
-    var filter = df.filter(df("reason").isNotNull.and(df("airport").endsWith("O")))
+    var filter = df.filter(df("reason").isNotNull.and(df("tag").equalTo("jan")))
```
danhermann, line 1047: Could you help me understand why `df("airport").endsWith("O")` changed to `df("tag").equalTo("jan")` here and on `sql-20/AbstractScalaEsSparkSQL.scala:1104`?

jbaiera: The endsWith predicate translates into a wildcard query. Usually we would lowercase the `O` character to make the query string be `*o`. Wildcard queries are term level queries and do not have analyzers applied to them. When run in "strict" mode, the query keeps the character uppercased (`*O`) which does not match the term in lucene. Lowercasing the `O` in the test would lead to Spark filtering the data out from the results (since `"o" != "O"`).

I swapped the endsWith predicate on the And-testing-methods because I wanted to bring back positive matches from the query in all test cases instead of assuming empty results when just strict mode is enabled.

jbaiera: more concisely: The bug deals with `strict` mode being enabled, and this test shouldn't be affected by that setting. Removing the feature that does depend on the setting from the test (hopefully) makes it simpler to reason about.

danhermann: Ah, ok, thanks for the explanation. LGTM!

jbaiera: Thanks!

PR #1686, opened 2021-06-10
Remove problem dependency

jbaiera:
Snapshot builds and CI is failing because a dependency cannot be reliably pulled from the repository it is in. The locations it is persisted in are either inaccessible or being sunset, and we luckily do not even use it, so this PR removes it from being considered by the build.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
@@ -172,7 +172,7 @@ class BuildPlugin implements Plugin<Project>  {
         project.repositories.mavenCentral()
         project.repositories.maven { url "https://conjars.org/repo" }
         project.repositories.maven { url "https://clojars.org/repo" }
-        project.repositories.maven { url 'https://repo.spring.io/plugins-release' }
+        project.repositories.maven { url 'https://repo.spring.io/plugins-release-local' }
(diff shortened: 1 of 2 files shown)


jbaiera:
Opened against master in error

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
```diff
@@ -172,7 +172,7 @@ class BuildPlugin implements Plugin<Project>  {
         project.repositories.mavenCentral()
         project.repositories.maven { url "https://conjars.org/repo" }
         project.repositories.maven { url "https://clojars.org/repo" }
-        project.repositories.maven { url 'https://repo.spring.io/plugins-release' }
+        project.repositories.maven { url 'https://repo.spring.io/plugins-release-local' }
```
mark-vieira, line 175: Why was this required _in addition_ to dropping the dependency?

mark-vieira (approved):

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/BuildPlugin.groovy
jbaiera: I've reverted this. It doesn't seem like we actually make use of this repository at all, especially since as it is it requires authorization to access. I'll put out a follow up to clean it out.

PR #1705, opened 2021-06-28
Fix buildSrc classpath after build-tools refactoring

breskeby:
We need to explicitly add the former transitive dependencies of commons codec and ant-rat to the buildSrc compile classpath as the code relying on this has moved directly to buildSrc here.

- [X] I have signed the [Contributor License Agreement (CLA)][]


jbaiera (approved):
LGTM

PR #1735, opened 2021-08-09
Update the development version of ES-Hadoop

jbaiera:
Fix the development version number for ES-Hadoop to be in line with the expected stack version for development.


mark-vieira (approved):
Good catch. LGTM.

PR #1744, opened 2021-09-07
Using case-insensitive headers

masseyke:
The check for the X-elastic-product header is currently case sensitive. However header names are
supposed to be case-insensitive. This commit makes the check case sensitive.
Relates #1745


mr/src/test/java/org/elasticsearch/hadoop/rest/RestClientTest.java
```diff
@@ -23,24 +23,14 @@
...
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
jbaiera, line 33: nit: Remove import-star's

jbaiera (approved):
LGTM, thanks for squashing this

mr/src/test/java/org/elasticsearch/hadoop/rest/RestClientTest.java
masseyke: Done.

Issue #1182, opened 2018-08-03
Add Rest Client Support for X-Opaque-ID


jbaiera:
Elasticsearch supports an HTTP header that is sent and returned on each request. Lately, support for this header (X-Opaque-Id) is getting more uses attached to it. Namely, the tasks api now references the opaque id of the request that led to the task, and the id now shows up in the node logs for Elasticsearch. We should make a point to generate meaningful values for this header and attach them to each request made to Elasticsearch.

masseyke:
I forgot to close this when the PRs went in, but this will be in 7.16.0 and 8.0.0.

PR #1755, opened 2021-09-15
Set X-Opaque-ID header in request header

masseyke:
Adding an X-Opaque-ID header to bulk requests from mapreduce, spark, and hive so that index requests from
elasticsearch-hadoop can be tracked in the tasks api, the slowlog, and the deprecations log.
Relates #1182

### DIFF
--- a/hive/src/main/java/org/elasticsearch/hadoop/hive/EsHiveOutputFormat.java
+++ b/hive/src/main/java/org/elasticsearch/hadoop/hive/EsHiveOutputFormat.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.ql.exec.FileSinkOperator.RecordWriter;
 import org.apache.hadoop.hive.ql.io.HiveOutputFormat;
+import org.apache.hadoop.hive.shims.ShimLoader;
 import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.util.Progressable;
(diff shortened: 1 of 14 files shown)


jbaiera:
Do we have a good idea of what is present in the task attempt id? I imagine that it's mostly book keeping information about the currently executing task, but not much about the job itself? I wonder if we should also see about pulling in the application name from the job settings for the opaque id as well?

jbaiera:
As for testing, I think we might be able to make this work with some unit testing primarily. Likely some test cases for each of the integrations with new code to make sure that they capture the task id/application info correctly, and a test for the client against some mocked http library objects to make sure it adds the header correctly?

masseyke:
> Do we have a good idea of what is present in the task attempt id? I imagine that it's mostly book keeping information about the currently executing task, but not much about the job itself? I wonder if we should also see about pulling in the application name from the job settings for the opaque id as well?

I believe for mapreduce jobs the taskAttemptId includes the job/application id (I need to confirm that though). The job name might also be useful. I'm not sure how verbose we want to get here, although there's only one of these per bulk request so I guess it won't take up that much bandwidth.
For spark the taskAttemptId is only a number like "3", so I included the applicationId as well.

jbaiera:
> I'm not sure how verbose we want to get here, although there's only one of these per bulk request so I guess it won't take up that much bandwidth.

I wonder if we should include the X-Opaque-Id on all requests sent to ES? The driver program can launch some searches that can easily get heavy in the planning phase that might be good to have tagged with client info.

jbaiera:
Should probably let users give their own opaque ids as well, override it with the job name if it's not set? Append the task id on the end of it if it's set.

masseyke:
Closing this one in favor of #1770, which is much simpler and more complete.

PR #1783, opened 2021-10-27
Fix script update tests in Spark

jbaiera:
Stored script request body has changed in 8.0. Updating test code to reflect this change.


masseyke (approved):
Looks good to me, although in order to even get to this failure I had to make several other changes (all of them were a result of the build going to java 17).

PR #1785, opened 2021-11-01
Build on java 17

masseyke:
This commit updates elasticsearch-hadoop to build on java 17. I did not look into it very much but Scala 2.10 and
Java 18 do not seem to be compatible. So I have removed support for Scala 2.10 as part of this. That also means
that support for Spark 1.x goes away altogether.
Relates #1791


masseyke:
I don't think we have any need to actually build on java 17 (yet) if we merge #1791.

masseyke:
Closing this in favor of the much smaller #1802

PR #1792, opened 2021-11-04
Getting rid of gradle deprecation warnings

masseyke:
Gradle has deprecated several features, and irritatingly it fails the build if you use those features without passing
the "--warning-mode=none" argument. This commit replaces those deprecated things so that you can run the build
without passing "--warning-mode=none".


jbaiera (approved):
Nice! LGTM assuming the build is happy

PR #1797, opened 2021-11-05
Backing up gradle to a version that works with scala 2.10

masseyke:
Gradle 7.2 causes failures building our scala 2.10 artifacts. This commit rolls it back to 7.1, which seems to work fine.


masseyke:
Here's an example build that fails with gradle 7.2:
```
./gradlew :elasticsearch-spark:compileSpark13scala210Scala
```
And here's an example of the kind of error we get:
```
[Error] : Class com.sun.beans.introspect.PropertyInfo not found - continuing with a stub.
```

PR #1800, opened 2021-11-05
Setting java boot classpath so that scala 2.10 works on gradle 7.2

masseyke:
Building scala 2.10 code stopped working when I upgraded to gradle 7.2 (scala 2.11 and 2.12 continued working
fine). I still don't fully understand it, but the problem seems to be that the java boot classpath jars are no longer
getting into the scala boot classpath. I found that I had to set it in two ways: (1) I have to explicitly set it in
scalaCompileOptions.additionalParameters or the build fails and (2) I have to set it in options.bootstrapClasspath
or I get warnings from javac that its boot classpath is incompatible with the source version. The two commands
I've been using to test this out are:
```
./gradlew :elasticsearch-spark:compileSpark13scala210Scala
```
and
```
./gradlew :elasticsearch-spark:test
```

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-bin.zip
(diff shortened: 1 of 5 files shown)


jbaiera (approved):
Looks good! Weird that it disappeared in 7.2

PR #1802, opened 2021-11-09
builds with java 17

masseyke:
This commit adds support for building on java 17, while leaving support for building java 11 untouched.


jbaiera (approved):
LGTM

PR #1803, opened 2021-11-09
Adding support for the date_nanos field type

### LINKED ISSUE #1653: Unable to read date_nanos
### Issue description

Unable to read date_nanos field to Spark df. 
Type date reading without problems but  date_nanos not show in df columns.

Strack trace:

No stack trace, code execute properly but don`t have  date_nanos column in results.

### Version Info

Hadoop/Spark: Spark 3.1.1
ES-Hadoop : elasticsearch-spark-30_2.12-7.12.0
ES : 7.7.1

masseyke:
Adding support for the date_nanos field type. This field type is now treated in the same way as the date field type.
Closes #1653

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/mr/WritableValueReader.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/mr/WritableValueReader.java
@@ -71,6 +71,7 @@ public class WritableValueReader extends JdkValueReader {
             arrayType = BooleanWritable.class;
             break;
         case DATE:
+        case DATE_NANOS:
             arrayType = dateType();
             break;
         case BINARY:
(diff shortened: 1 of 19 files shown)


jbaiera (commented):
Does the code suffer from any loss of precision issues from using nano time on older time libraries? Can we add a test to make sure the serialization for date nanos works?

masseyke:
> Does the code suffer from any loss of precision issues from using nano time on older time libraries? Can we add a test to make sure the serialization for date nanos works?

You're right -- the parser only produces milliseconds, and we're only grabbing milliseconds from it at https://github.com/masseyke/elasticsearch-hadoop/blob/feature/date-nanos/spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueReader.scala#L181.

masseyke:
OK, I've changed it so that for date_nanos fields we keep track of the nanoseconds the whole way through. In addition to the unit tests, I manually pulled data out of an index in spark shell and wrote it out to a new index, and the nanoseconds were still correctly there.
I originally converted regular date fields to use java time instead of the mix of java 6 and joda time (uncommitted), but decided to revert that as out of scope for this PR. This should only impact the path taken for date_nanos. We probably ought to switch everything over at some point, now that we only support java 8 and newer.

jbaiera (approved):
LGTM

PR #1807, opened 2021-11-17
Add support for Spark 3.1 and 3.2

jbaiera:
Spark changed the internal implementation of the CompactibleFileStreamLog which caused issues when trying to upgrade. We simply support both the old and new version by implementing both required methods (old and new) and allowing them to be resolved at runtime as needed.


gradle.properties
```diff
@@ -30,7 +30,7 @@ spark13Version = 1.6.2
 spark20Version = 2.3.0
 spark22Version = 2.2.3
 spark24Version = 2.4.4
-spark30Version = 3.0.1
+spark30Version = 3.2.0
```
masseyke, line 33: This is a minor thing, but should we call this spark32Version, or spark3Version? Seems odd (and inconsistent with what we've done in spark 2.x) to have the spark 3.0 version be 3.2.

masseyke:
@elasticmachine update branch

gradle.properties
```diff
@@ -30,7 +30,7 @@ spark13Version = 1.6.2
 spark20Version = 2.3.0
 spark22Version = 2.2.3
 spark24Version = 2.4.4
-spark30Version = 3.0.1
+spark32Version = 3.2.0
```
masseyke, line 33: This property rename seems to have broken the build (I think there were other places it was referenced. I'm not sure it's worth it. I changed these back locally, and everything seems to work for me.

jbaiera: I reverted it - I think keeping it `30` is better for tracking earliest compatibility in the 3.x line for this version

masseyke (approved):
Looks good to me. I built it out and tried it manually in a yarn cluster -- didn't find any problems.

olegbonar:
Why `elasticsearch-spark-30_2.12:8.0.0` still depends on `org.apache.spark:spark-catalyst_2.12:3.0.1`? Shouldn't it depend on spark 3.2.1 already?

masseyke:
It looks like a mistake. It looks like we're pulling in both 3.0.1 and 3.2.0. I believe the 3.2.0 ones are taking precedence (b/c we're able to use 3.2-only features). But we need to sort that out.

PR #1808, opened 2021-12-01
Build Hadoop with Java 17

mark-vieira:
We are bumping the compiler version of [Elasticsearch to Java 17](https://github.com/elastic/elasticsearch/pull/80851) meaning that `build-tools` will now target Java 17. For this not to break elasticsearch-hadoop, we need to start building it with Java 17 as well.


masseyke (approved):

masseyke:
Relates #1802

PR #1827, opened 2021-12-20
Fixing support for empty and null arrays

### LINKED ISSUE #1527: array type default is array and has one null value

masseyke:
This commit makes it so that if an array is null in Elasticsearch, it comes into es-hadoop/spark as null. Previously
it showed up as either [] or [null], depending on spark version.
Closes #1527

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/serialization/ScrollReader.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/serialization/ScrollReader.java
@@ -895,7 +895,12 @@ public class ScrollReader implements Closeable {
             String rawValue = parser.text();
             try {
                 if (isArrayField(fieldMapping)) {
-                    return singletonList(fieldMapping, parseValue(parser, esType), parser);
+                    Object parsedValue = parseValue(parser, esType);
+                    if (parsedValue == null) {
+                        return null; //There is not a null element in the array. The array itself is null.
+                    } else {
+                        return singletonList(fieldMapping, parsedValue, parser);
+                    }
                 } else {
                     return parseValue(parser, esType);
                 }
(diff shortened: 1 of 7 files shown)


jbaiera (approved):
LGTM

PR #1839, opened 2021-12-29
Adding support for update output mode to structured streaming

### LINKED ISSUE #1123: Support UPDATE output mode for Spark Structured Streaming
We now support Spark Structured Streaming as of 6.x, but only in `APPEND` mode. It would be beneficial to support `UPDATE` mode, with the requirement that the user MUST provide a field to be used for a document ID.

masseyke:
This commit adds support for "update" as the output mode for spark structured streaming to Elasticsearch.
Closes #1123

### DIFF
--- a/spark/sql-20/src/itest/scala/org/elasticsearch/spark/sql/streaming/StreamingQueryTestHarness.scala
+++ b/spark/sql-20/src/itest/scala/org/elasticsearch/spark/sql/streaming/StreamingQueryTestHarness.scala
@@ -76,7 +76,7 @@ class StreamingQueryTestHarness[S <: java.io.Serializable : Encoder](val sparkSe
     private var foundExpectedException: Boolean = false
     private var encounteredException: Option[String] = None
 
-    private val latch = new CountDownLatch(1)
+    private var latch = new CountDownLatch(1) // expects just a single batch
 
     def incrementExpected(): Unit = inputsRequired = inputsRequired + 1
 
@@ -153,6 +153,10 @@ class StreamingQueryTestHarness[S <: java.io.Serializable : Encoder](val sparkSe
 
     def waitOnComplete(timeValue: TimeValue): Boolean = latch.await(timeValue.millis, TimeUnit.MILLISECONDS)
 
+    def expectAnotherBatch(): Unit = {
+      latch = new CountDownLatch(1)
+    }
+
     def assertExpectedExceptions(message: Option[String]): Unit = {
       expectingToThrow match {
         case Some(exceptionClass) =>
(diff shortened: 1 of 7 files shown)


masseyke:
I'm new to structured streaming, but I think that this does what we'd expect. That is, if you set output mode to "update", then it inserts documents if they didn't exist before, and only updates documents that have changed. That's what my understanding from https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#basic-concepts is of the expected behavior. In addition to the integration test that's part of this PR, I did a manual test (using the docker image at https://github.com/masseyke/es-spark-docker)
```
import org.apache.spark.sql.SparkSession
case class Person(id: String, name: String, surname: String, age: Int)
val people = spark.readStream.textFile("/test/*").map(_.split(",")).map(p => Person(p(0), p(1).trim, p(2).trim, p(3).trim.toInt))
val appendStream = people.writeStream.outputMode("append").option("checkpointLocation", "/save/location").option("es.mapping.id", "name").format("es").start("people")
```
Then I put a 4-row, 4-col csv file in hdfs at /test/, and made sure those 4 rows showed up as documents in Elasticsearch. And back in spark shell:
```
appendStream.stop
```
And then I added a second csv file to /test/ in HDFS that had the same rows as the first csv file but one of them had an updated value. Then back in spark shell:
```
val updateStream = people.writeStream.outputMode("update").option("checkpointLocation", "/save/location").option("es.mapping.id", "name").format("es").start("people")
```
Then with a query to Elasticsearch that returns version (GET people/_search?version=true) I verified that only the one document had been updated.
As @jbaiera mentioned in #1123, I'm intentionally blowing up if an "es.mapping.id" is not provided since we can't really update otherwise. Also I'm intentionally blowing up if "es.write.operation" is set to something other than "upsert" since based on the spark documentation upsert is what we want here. If it is unset, I set it to upsert.

spark/sql-20/src/itest/scala/org/elasticsearch/spark/sql/streaming/StreamingQueryTestHarness.scala
```diff
@@ -320,6 +324,30 @@ class StreamingQueryTestHarness[S <: java.io.Serializable : Encoder](val sparkSe
     }
   }
 
+  /**
+   * Waits until all inputs are processed on the streaming query, but leaves the query open with the listener still in place, expecting
+   * another batch of inputs.
+   */
+  def waitForPartialCompletion(): Unit = {
```
jbaiera, line 331: Nice!

jbaiera (approved):
LGTM!

PR #1478, opened 2020-05-16
resolve saveToEs saves case classes fields with NULL values #998

lucebert:
This PR fix the Scala Writer when it processes case classes with null fields. 

Correction about issue #998  

- [x] I have signed the [Contributor License Agreement (CLA)][]

PR #1855, opened 2022-01-06
Fixing a scala compilation problem

masseyke:
The code in ScalaValueWriter::doWriteScala was not compiling on the 7.17 branch. The problem is that the () reference to Unit was confusing the compiler. This commit makes the reference to Unit more explicit, and adds a test to make sure that Unit is being handled as expected.

### DIFF
--- a/spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
+++ b/spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
@@ -61,7 +61,14 @@ class ScalaValueWriter(writeUnknownTypes: Boolean = false) extends JdkValueWrite
         generator.writeBeginObject()
         for ((k, v) <- m) {
           if (shouldKeep(parentField, k.toString)) {
-            if (v != None && v!= null && v != () || hasWriteNullValues) {
+            val nonEmptyValue = Option(v) match {
+              case Some(()) => false
+              case Some(None) => false
+              case Some(_) => true
+              case None => true
+              case _ => false
+            }
+            if (nonEmptyValue || hasWriteNullValues) {
               generator.writeFieldName(k.toString)
               val result = doWrite(v, generator, k.toString)
               if (!result.isSuccesful) {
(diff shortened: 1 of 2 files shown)


masseyke:
Specifically, this was failing on the 7.17 branch:
```
./gradlew :elasticsearch-spark-13:assemble
```

spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
```diff
@@ -61,7 +61,14 @@ class ScalaValueWriter(writeUnknownTypes: Boolean = false) extends JdkValueWrite
         generator.writeBeginObject()
         for ((k, v) <- m) {
           if (shouldKeep(parentField, k.toString)) {
-            if (v != None && v!= null && v != () || hasWriteNullValues) {
+            val nonEmptyValue = Option(v) match {
+              case Some(()) => false
+              case Some(None) => false
+              case Some(_) => true
+              case None => true
+              case _ => false
```
jbaiera, line 69: Is the catch all needed here? This confused me for a second

spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
```diff
@@ -61,7 +61,14 @@ class ScalaValueWriter(writeUnknownTypes: Boolean = false) extends JdkValueWrite
         generator.writeBeginObject()
         for ((k, v) <- m) {
           if (shouldKeep(parentField, k.toString)) {
-            if (v != None && v!= null && v != () || hasWriteNullValues) {
+            val nonEmptyValue = Option(v) match {
```
jbaiera, line 64:
```suggestion
            val hasValue = Option(v) match {
```

Can we swap the variable name? `nonEmptyValue=false` reads weird in my head

spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
```diff
@@ -61,7 +61,14 @@ class ScalaValueWriter(writeUnknownTypes: Boolean = false) extends JdkValueWrite
...
+              case Some(()) => false
+              case Some(None) => false
+              case Some(_) => true
+              case None => true
+              case _ => false
+            }
+            if (nonEmptyValue || hasWriteNullValues) {
```
jbaiera, line 71:
```suggestion
            if (hasValue || hasWriteNullValues) {
```
If applying above suggestion

jbaiera (approved):
LGTM, minor nit and optional simplification

spark/core/src/main/scala/org/elasticsearch/spark/serialization/ScalaValueWriter.scala
masseyke: Oops good point!

masseyke:
Relates #1478

PR #1870, opened 2022-01-14
Checking result of bulk writes

masseyke:
AbstractBulkFactory currently ignores the results of certain writes, meaning that a user might corrupt or lose data without realizing it. This commit checks the result and explicitly throws an Exception if there is a problem.


jbaiera (approved):
🚀

PR #1871, opened 2022-01-19
Feature/enable kerberos tests

### LINKED ISSUE #1366: Re-enable QA Kerberos Tests
The QA Kerberos tests were disabled immediately after they were added earlier in the year due to some issues with their compatibility with our CI environment. First and foremost, there are a set of jobs that fail due to path resolution errors: Hadoop incorrectly treats + signs in paths as spaces, and thus, file paths under the project in CI are broken. See #1251 .

Additionally, even getting the jobs to respect path names correctly in CI still encounters failures. The YARN environment needs to be tuned for the lower amount of memory available in the CI environment. Right now, the default settings are far too greedy and need to be reigned in.

Finally, after retrying to run them in a local test environment, it seems that auth issues are starting to occur due to local host name resolution on Kerberos. Further investigation is required into these failures.

Relates #1251

masseyke:
This commit re-enables the kerberos tests that had been disabled in https://github.com/elastic/elasticsearch-hadoop/commit/b3547fc42785ca071583178d1e4317262772435e.
Closes #1366

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
@@ -22,11 +22,15 @@ package org.elasticsearch.hadoop.gradle.fixture.hadoop.tasks
 import org.elasticsearch.gradle.testclusters.DefaultTestClustersTask
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfiguration
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.InstanceConfiguration
+import org.gradle.api.tasks.Internal
 
 abstract class AbstractClusterTask extends DefaultTestClustersTask {
 
+    @Internal
     HadoopClusterConfiguration clusterConfiguration
+    @Internal
     InstanceConfiguration executedOn
+    @Internal
     Map<String, String> environmentVariables = [:]
 
     AbstractClusterTask() {
@@ -41,6 +45,7 @@ abstract class AbstractClusterTask extends DefaultTestClustersTask {
     abstract InstanceConfiguration defaultInstance(HadoopClusterConfiguration clusterConfiguration)
     abstract Map<String, String> taskEnvironmentVariables()
 
+    @Internal
     protected getInstance() {
         return executedOn == null ? defaultInstance(this.clusterConfiguration) : executedOn
     }
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
@@ -22,15 +22,17 @@ package org.elasticsearch.hadoop.gradle.fixture.hadoop.tasks
 import org.elasticsearch.gradle.testclusters.DefaultTestClustersTask
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfiguration
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.InstanceConfiguration
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Optional
 
 abstract class AbstractClusterTask extends DefaultTestClustersTask {
 
-    @Internal
+    @Input
     HadoopClusterConfiguration clusterConfiguration
-    @Internal
+    @Input @Optional
     InstanceConfiguration executedOn
-    @Internal
+    @Input
     Map<String, String> environmentVariables = [:]
 
     AbstractClusterTask() {
@@ -45,7 +47,7 @@ abstract class AbstractClusterTask extends DefaultTestClustersTask {
     abstract InstanceConfiguration defaultInstance(HadoopClusterConfiguration clusterConfiguration)
     abstract Map<String, String> taskEnvironmentVariables()
 
-    @Internal
+    @Input
     protected getInstance() {
         return executedOn == null ? defaultInstance(this.clusterConfiguration) : executedOn
     }
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
@@ -23,7 +23,6 @@ import org.elasticsearch.gradle.testclusters.DefaultTestClustersTask
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfiguration
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.InstanceConfiguration
 import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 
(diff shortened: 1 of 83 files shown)


masseyke:
This doesn't actually make the changes described in #1366. I want to see if they're still relevant before doing that (a lot has changed in 3 years, including the CI system).

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
```diff
@@ -22,11 +22,15 @@ package org.elasticsearch.hadoop.gradle.fixture.hadoop.tasks
...
 import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.InstanceConfiguration
+import org.gradle.api.tasks.Internal
 
 abstract class AbstractClusterTask extends DefaultTestClustersTask {
 
+    @Internal
     HadoopClusterConfiguration clusterConfiguration
+    @Internal
     InstanceConfiguration executedOn
+    @Internal
```
jbaiera, line 33: `@Internal` is used to indicate that the marked task property changing shouldn't cause a task to be rerun, but these all seem to be pretty important items - if the environment changes we should probably run. I'm not sure on the cluster config or which instance to execute from, but my gut says those as well. Should these all be `@Input` instead?

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/HadoopMRJob.groovy
```diff
@@ -23,18 +23,25 @@ import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfigur
...
 
 import static org.elasticsearch.hadoop.gradle.util.ObjectUtil.unapplyString
 
 class HadoopMRJob extends AbstractClusterTask {
 
+    @Internal
     String jobClass
+    @Internal
     File jobJar
+    @Internal
```
jbaiera, line 38: Same question re: `@Input` for these as well

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/HiveBeeline.groovy
```diff
@@ -23,15 +23,19 @@ import org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.HadoopClusterConfigur
...
 import org.gradle.api.GradleException
+import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.TaskAction
 import org.gradle.process.ExecSpec
 
 import static org.elasticsearch.hadoop.gradle.fixture.hadoop.conf.SettingsContainer.FileSettings
 
 class HiveBeeline extends AbstractClusterTask {
 
+    @Internal
```
jbaiera, line 34: Same re: `@Input`

We should probably change this file type to a path type since I saw that Gradle is weird about File types for Inputs

qa/kerberos/src/main/resources/test.properties
```diff
@@ -0,0 +1,57 @@
...
+# put pressure on the bulk API
+es.batch.size.entries=3
+es.batch.write.retry.wait=1s
+
+mapred.max.split.size=134217728
+#mapreduce.local.map.tasks.maximum=10
+
+# M&R
+fs.default.name=file:///
+mapred.job.tracker=local
```
jbaiera, line 27: Do these properties cause the job runner to execute in local mode instead of on the test cluster that is set up?

qa/kerberos/src/main/resources/test.properties
```diff
@@ -0,0 +1,57 @@
...
+
+# Locally Deployed MR
+#fs.defaultFS=hdfs://localhost:9000
+#mapreduce.framework.name=yarn
+
+#Hive - only works if the HDFS is setup as well
+hive=local
```
jbaiera, line 34: Same question for Hive re: Local mode

qa/kerberos/src/main/resources/test.properties
```diff
@@ -0,0 +1,57 @@
...
+#es.net.proxy.http.pass=test
+
+# Pig
+pig.script.info.enabled=false
+
+# Storm
+es.storm.bolt.flush.entries.size=1
+
+# Spark props
+spark.master=local
```
jbaiera, line 53: re: Local mode?

jbaiera (commented):
Had some questions about the usage of the Gradle annotations and around the new test properties file added

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
masseyke: Hmm maybe I misread the documentation for that annotation. You're probably right. I thought that if it was `@Internal` it would treat it as something that needed to be computed every time. I had intended to revisit it to see if I could use `@Input` here to speed things up if nothing had changed.

qa/kerberos/src/main/resources/test.properties
masseyke: Actually good question. I _thought_ that these were the properties that we had been using. I copied them from test/shared/src/main/resources because (I assumed) that something had changed in the classpath over the years, and [loading](https://github.com/elastic/elasticsearch-hadoop/blob/master/test/shared/src/main/java/org/elasticsearch/hadoop/util/TestSettings.java#L41) had stopped finding that properties file on the classpath. It's possible that there was another test.properties that has been deleted? It looks like that class sets those two properties from system properties though.

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/HiveBeeline.groovy
masseyke: Have a reference for the weirdness? This seems to make it sound like it's OK as long as it's not a relative File, right? https://docs.gradle.org/current/userguide/working_with_files.html#sec:locating_files

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/AbstractClusterTask.groovy
masseyke: Yeah I had misread this. Switched them over to @Input (except for a couple that are never actually changed as far as I can tell).

qa/kerberos/src/main/resources/test.properties
masseyke: I don't think any of these were actually being used. I've removed them, and confirmed that when I run the tests I see all of the jobs in the resource manager (http://localhost:8088) so I think we're good here.

qa/kerberos/src/main/resources/test.properties
masseyke: Yeah, it's definitely running in yarn mode (see above).

buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/tasks/HiveBeeline.groovy
jbaiera: I based it off of the `@Input` annotation docs. https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Input.html

jbaiera: Ah I see you're using `@InputFile` instead. All good!

jbaiera (approved):
LGTM if CI is happy! Woo!

Issue #69, opened 2013-08-14
Index aware document writes


costin:
Currently elasticsearch-hadoop supports only document creation (id is always generated by ES).

This area needs to be improved to support the following strategies:
1. create/put-if-absent (applies also when there's no id)
2. index op-type=index or create document if missing, _replace_ if present
3. upsert or create document if missing, _update if present

PR #1878, opened 2022-01-24
Expected exception when incorrectly configuring upsert has changed

masseyke:
AbstractMROldApiSaveTest.testUpdateWithoutId broke when #1839 was merged because we are now failing earlier if
upsert is used but `es.mapping.id` is not set. The exception is now the same as the one you get when update is not
configured to use `es.mapping.id`.
Relates #1839 #69


masseyke:
I went ahead and merged this one since the build is broken and @jbaiera and I have discussed it outside of the ticket.

PR #1895, opened 2022-02-02
[DOCS] Add latest bug fix to 8.0.0-rc1 release notes

lockewritesdocs:
Adds an entry for #1870 to the 8.0.0-rc1 release notes, which was missing.


lockewritesdocs:
@masseyke, can you provide a 👍 before I merge this?

masseyke:
This is not in the 8.0 branch -- it's only in master.

masseyke (changes requested):
Maybe this was supposed to be some other PR? I'll look through the list, but I intentionally never backported #1870 to 8.0 because I wanted a little more time than we have for 8.0 to see if it would cause any unexpected problems.

lockewritesdocs:
@masseyke, you're right on this one -- @pugnascotia thought that there might be a missing release note for `rc1`, but the PR that I included was only for 8.1.0. I'm going to close this without merging.

PR #1899, opened 2022-02-03
Updating IOUtils.toCanonicalFilePath to work in a spring boot application

### LINKED ISSUE #1359: Version check fails on SpringBoot applications
It seems that the Spring Boot plugin includes the dependencies of the jars in your project nested inside the final jar. This is causing a "hierarchical URI exception."

When ES-Hadoop is first loaded, it runs some checks within the Version class that ensure that only one copy/version of ES-Hadoop is available on the classloader. This is to ensure no Jar-Hell issues while running (which was more common than one would think without the check).

To verify that condition, the ES-Hadoop asks the current classloader for all copies of the Version class that it knows about. The classloader returns a set of URL's (either jars or .class files) that are then validated by ES-Hadoop to ensure no duplicates. We canonicalize each URL to eliminate false duplicates, like if a jar appears twice because of symlinks in the classpath. Finally, if only one unique URL remains, we can be sure that there's only one ES-Hadoop library available and the check passes.

The issue here is that the location of the jar file that ES-Hadoop finds is within another jar file, which cannot be resolved as a canonical filepath because it is inside an archive file. In this case, the ES-Hadoop jar is located at:
```
file:///path/to/agensspark-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/elasticsearch-hadoop-7.3.2.jar
```
... which is not a hierarchical URI.

masseyke:
This commit makes it so that es-hadoop does not blow up when it is packaged within a Spring Boot application.
Closes #1359

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/util/IOUtils.java
@@ -33,6 +33,8 @@ import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
+import java.net.URLStreamHandler;
 import java.util.Properties;
 
 import javax.xml.bind.DatatypeConverter;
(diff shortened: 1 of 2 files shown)


masseyke:
I cloned the project at https://github.com/maxmin93/agens-spark to test this. I realized that the problem is more than just that it's a jar with in a jar -- it is also complicated by the fact that Spring Boot provides its own custom URLStreamHandler. That URLStreamHandler results in the URLConnection returning a different jar file URL than the default one. I added a unit test to simulate what the spring boot one does.

jbaiera (approved):
LGTM

PR #1906, opened 2022-02-09
Getting rid of gradle warnings

masseyke:
This commit adds a dependency between the top-level javadoc task and the sub-level javadoc tasks so that we don't get
gradle 8 warnings.


jbaiera (commented):
Very strange... do we need that root javadoc task still?

masseyke:
If we don't have it, then we don't get the javadocs in the dist directory. I assume we need them there? Although I don't actually know how (or if) we're distributing javadocs.

jbaiera:
We are distributing javadocs I believe. You sure we don't get any? I just commented out the `configureProjectJars()` function in `IntegrationBuildPlugin` and they still generated under `dist/build/distributions` for me. I think the `IntegrationBuildPlugin` might have survived a round of cleanup in the project build code.

masseyke:
I was running `./gradlew :assemble`. Will check again.

masseyke:
@jbaiera if I comment out `configureProjectJars` then when I run `./gradlew :assemble` then there is nothing in `build/docs/javadoc/`. Maybe I'm misunderstanding you?

masseyke:
Ha I was definitely looking at the wrong thing then! I'll just remove that call and update this PR.

jbaiera (approved):
LGTM - We may want to open a TODO issue for cleaning up the `IntegrationBuildPlugin` and `RootBuildPlugin` classes.

PR #1919, opened 2022-02-23
Backporting support for gradle for 7.3

masseyke:
This commit backports #1800 and #1809 in order to support gradle 7.3 since gradle 7.3 is required in order
to build against elasticsearch 7.17.1 due to https://github.com/elastic/elasticsearch/pull/84023.


jbaiera (approved):
LGTM

PR #1924, opened 2022-03-02
Adding an explicit dependency on jar to setupUsers

masseyke:
It looks like `:qa:kerberos:setupUsers` fails on a clean installation because it has an implicit dependency on the :qa:kerberos code having been built. This commit adds an explicit dependency. I happened to stumble on this while tracking down other problems, but I don't think it impacts any of our normal builds.


jbaiera (approved):
LGTM

PR #1934, opened 2022-03-28
Set UserProvider before discovery in Spark SQL integrations

### LINKED ISSUE #1933: SparkSQL breaks when saving data with Kerberos authentication
When submitting a job on ES-Hadoop, one of the initialization steps that needs to happen before the connector tries contacting Elasticsearch is to set the UserProvider implementation to be used. The UserProvider implementation establishes the logic for persisting credentials and performing actions as users based on what the current integration supports.

There is currently an issue with the Spark SQL integration surrounding its initialization code. In some places it fails to set any user provider class before performing the initial cluster discovery. In other places it sets the provider, but only after attempting discovery. Not having the provider configured before discovery leads to no credentials being passed to the underlying HTTP client which leads it to attempt a discovery with no API key. The Elasticsearch server rejects the request because it is unauthenticated and the spark job will fail.

A workaround for this bug is to specify the UserProvider class in your spark driver when saving the dataset or data frame:

```scala
dataframe.saveToEs("index_name", Map("es.security.user.provider.class" -> "org.elasticsearch.hadoop.mr.security.HadoopUserProvider"))
```

Alternatively, when submitting with the data source api, the setting can be provided with the following options methods:
```scala
dataframe.write
  .format("es")
  .option("es.security.user.provider.class", "org.elasticsearch.hadoop.mr.security.HadoopUserProvider")
  .save("index_name")
```

jbaiera:
This PR adds the requisite configuration step for the UserProvider implementation before attempting to discover the cluster information in the SparkSQL integrations. This additionally updates the integration tests to add coverage for the write and read paths for SparkSQL.

Fixes #1933

### DIFF
--- a/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
+++ b/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
@@ -151,6 +151,7 @@ private[sql] case class ElasticsearchRelation(parameters: Map[String, String], @
 
   @transient private[sql] lazy val cfg = {
     val conf = new SparkSettingsManager().load(sqlContext.sparkContext.getConf).merge(parameters.asJava)
+    InitializationUtils.setUserProviderIfNotSet(conf, classOf[HadoopUserProvider], LogFactory.getLog(classOf[ElasticsearchRelation]))
     InitializationUtils.discoverClusterInfo(conf, LogFactory.getLog(classOf[ElasticsearchRelation]))
     conf
   }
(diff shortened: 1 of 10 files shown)


spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
```diff
@@ -533,10 +534,10 @@ private[sql] case class ElasticsearchRelation(parameters: Map[String, String], @
 
       // perform a scan-scroll delete
       val cfgCopy = cfg.copy()
+      InitializationUtils.setUserProviderIfNotSet(cfgCopy, classOf[HadoopUserProvider], null)
```
masseyke, line 537: Should we just roll `setUserProviderIfNotSet` into `discoverClusterInfo` to keep us from accidentally calling `discoverClusterInfo` without it in the future?

masseyke (approved):
I have the one question about whether we could do something to prevent this kind of thing in the future, but LGTM.

spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala
jbaiera: I don't disagree, but I think the problems are rooted deeper than just this initialization call. Honestly, I think all of `InitializationUtils` in general needs to be put on a bus.

Issue #1140, opened 2018-04-23
Support for signing requests for AWS Elasticsearch


prashs:
### What kind an issue is this?

- [x] Feature Request. 

### Issue description
If we want to fetch or write data to/from an AWS Elasticsearch setup, there is a need to sign requests as per the below documentation:

https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-ac.html#es-managedomains-signing-service-requests

As far as I understand, this is not supported by the "elasticsearch-hadoop" library as of now. If spark is running on AWS, one could configure IP address/host based IAM policies, but if data needs to be accessed from outside of AWS, there is not a good solution for this as far as I know.

The ES proxy (https://github.com/abutaha/aws-es-proxy) might be a possible way to get around this, but the problem there is that the infrastructure teams usually hesitate to provide tools that anybody can access without authenticating and dev teams don't really want to maintain things like this proxy.

### Version Info

OS:         :  Centos 6.8
JVM         :  1.8
Hadoop/Spark: Spark 2  
ES-Hadoop   :  6.2.4
ES          :  AWS ES 5.5

### Feature description
See description above

jbaiera:
Hi @prashs, thanks for opening an issue for this. We currently have an issue that covers this problem at #626. If you would like to add any information on to that issue as well as any +1's that would be great.

Since we have an issue open for this already, I'll go ahead and close this. Let me know if there's any reasons you feel like this should be re-opened and I'd be happy to take a look.

Issue #565, opened 2015-10-07
Is it possible to write to Amazon Elasticsearch Service using elasticsearch-hadoop?


fblundun:
I have been trying to write to AWS's new Amazon Elasticsearch Service from a Scalding job using elasticsearch-hadoop (via [scalding-taps](https://github.com/scalding-io/scalding-taps)).

This job has previously worked using Elasticsearch manually installed on an EC2 instance, implying that the problem is specific to Amazon Elasticsearch Service.

The first time I tried I got this error:

```
Caused by: org.elasticsearch.hadoop.EsHadoopIllegalStateException: Cluster state volatile; cannot find node backing shards - please check whether your cluster is stable
        at org.elasticsearch.hadoop.rest.RestRepository.getWriteTargetPrimaryShards(RestRepository.java:370)
        at org.elasticsearch.hadoop.rest.RestService.initSingleIndex(RestService.java:425)
        at org.elasticsearch.hadoop.rest.RestService.createWriter(RestService.java:393)
        at org.elasticsearch.hadoop.mr.EsOutputFormat$EsRecordWriter.init(EsOutputFormat.java:173)
        at org.elasticsearch.hadoop.mr.EsOutputFormat$EsRecordWriter.write(EsOutputFormat.java:149)
        at org.apache.hadoop.mapred.MapTask$DirectMapOutputCollector.collect(MapTask.java:844)
        at org.apache.hadoop.mapred.MapTask$OldOutputCollector.collect(MapTask.java:596)
        at cascading.tap.hadoop.util.MeasuredOutputCollector.collect(MeasuredOutputCollector.java:69)
        at org.elasticsearch.hadoop.cascading.EsHadoopScheme.sink(EsHadoopScheme.java:212)
        at cascading.tuple.TupleEntrySchemeCollector.collect(TupleEntrySchemeCollector.java:153)
        ... 21 more
```

I then tried setting `es.nodes.client.only` to `true` and instead got this error:

```
Caused by: org.elasticsearch.hadoop.EsHadoopIllegalArgumentException: Client-only routing specified but no client nodes with HTTP-enabled were found in the cluster...
        at org.elasticsearch.hadoop.rest.InitializationUtils.filterNonClientNodesIfNeeded(InitializationUtils.java:82)
        at org.elasticsearch.hadoop.rest.RestService.createWriter(RestService.java:373)
[...]

PR #1941, opened 2022-04-04
Add Support AWS Signature V4

lucasvieirasilva:
This PR adds support for AWS Signature V4.

This feature helps AWS users to use IAM credentials instead of using basic authentication only or using HTTP Proxies to sign the requests.

#### Related Issues

- #565 
- #1140

#### Implementation

Because this project still uses the HTTP client `commons-httpclient:commons-httpclient:3.0.1` which does not support interceptors, the default AWS recommendation https://docs.aws.amazon.com/opensearch-service/latest/developerguide/request-signing.html#request-signing-java is not supported.

To solve the issue I had to implement the Sign V4 logic in this PR as well.

I have tested locally and the signature request is being signed and correctly authenticated in the AWS service.

The package size increased from `2mb` to `5.2mb`.

I also used only the AWS SDK Code module, which includes only the core modules, enough for this requirement.


jbaiera:
Thanks for the contribution but for a number of reasons I don't think we can support this. I do not think we will be able to support running on AWS Elasticsearch due to lack of product headers in responses, plus supporting another distribution becomes much more difficult in regards to testing (I see that there are no tests currently either). I do really appreciate the time you've put into this but I'm afraid we have to decline the PR.

stanpalatnik:
@jbaiera This has nothing to do with "supporting another distribution". People are already using these libraries to talk to AWS Opensearch if the authentication is disabled. The only change here is adding support for *signing* the requests to the clusters that this library already implicitly supports anyway. It's clear to me that closing this MR is purely a political decision because of the ongoing issues with Elasticsearch vs Opensearch. There is absolutely no justification for rejecting this MR outside of it.

PR #1943, opened 2022-04-06
Fixing DST bug in data_nanos support

masseyke:
In #1803 we added support for the `data_nanos` Elasticsearch field type. However there is a bug in that sets the zone offset to the offset right now, as opposed to what the zone offset was at the time of the date. So for example if the date was `2015-01-01T06:10:30.123456789-06:00`, then if you read that date after daylight saving had begun you would see it as `2015-01-01T06:10:30.123456789-05:00`.


jbaiera (approved):
LGTZ

PR #1944, opened 2022-04-06
Updating the Hadoop fixture version to 3.3.2

masseyke:
It looks like 3.3.0 is no longer available for download at https://dlcdn.apache.org/hadoop/common/, so updating to the latest.


jbaiera (approved):
LGTM

PR #1950, opened 2022-04-18
Backport java 17 build support to 7.17

masseyke:
Backporting support for building on java 17 from #1802 and #1808

### DIFF
--- a/hive/build.gradle
+++ b/hive/build.gradle
@@ -59,9 +59,10 @@ dependencies {
 }
 
 tasks.named("test").configure {
-    onlyIf { JavaVersion.current() >= JavaVersion.VERSION_16 }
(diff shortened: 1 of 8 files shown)


hive/build.gradle
```diff
@@ -58,6 +58,12 @@ dependencies {
     javadocSources(project(":elasticsearch-hadoop-mr"))
 }
 
+tasks.named("test").configure {
+    onlyIf { JavaVersion.current() >= JavaVersion.VERSION_16 }
```
jbaiera, line 62: Is this just needed to conditionally add the jvmArgs settings or are these tests required to run on JVM 16 and up?

masseyke: This is just for the jvmArgs settings. The tests still run for all versions right?

jbaiera: I think the `onlyIf` method marks the task to only run when the condition is true

jbaiera (approved):
LGTM, thanks!

PR #1982, opened 2022-07-25
Update Spark QA version

jbaiera:
Updates the Spark QA tarball version to latest in its minor version.


masseyke (approved):
Any reason to do this and not upgrade the version we build with in gradle.properties? LGTM though.

Issue #1992, opened 2022-08-23
Start emitting the User-Agent and X-Elastic-Client-Meta HTTP headers


sethmlarson:
(Content to be added here)

PR #1993, opened 2022-08-24
Adding a user-agent header

masseyke:
Adding a simple user agent that contains just es-hadoop version information and hadoop or spark version information. For example:
```
elasticsearch-hadoop/8.5.0-SNAPSHOT spark/1.6.2
```

### DIFF
--- a/spark/core/src/main/scala/org/elasticsearch/spark/cfg/SparkSettings.java
+++ b/spark/core/src/main/scala/org/elasticsearch/spark/cfg/SparkSettings.java
@@ -25,7 +25,6 @@ import java.util.Properties;
 
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.spark.SparkConf;
-import org.apache.spark.SparkContext;
 import org.elasticsearch.hadoop.cfg.Settings;
 import org.elasticsearch.hadoop.util.Assert;
 import org.elasticsearch.hadoop.util.IOUtils;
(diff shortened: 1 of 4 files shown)


spark/core/src/main/scala/org/elasticsearch/spark/cfg/SparkSettings.java
```diff
@@ -51,6 +53,13 @@ public SparkSettings(SparkConf cfg) {
         String appId = cfg.get("spark.app.id", "");
         String opaqueId = String.format(Locale.ROOT, "[spark] [%s] [%s] [%s]", user, appName, appId);
         this.setOpaqueId(opaqueId);
+        this.setUserAgent(String.format(
+                        Locale.ROOT,
+                        "elasticsearch-hadoop/%s spark/%s",
+                        Version.versionNumber(),
+                        SparkContext.getOrCreate(cfg).version()
```
jbaiera, line 60: Is there a lighter way to get this version or are we sure that the SparkContext will already be present when getting it? Would creating the context here and not closing it out be problematic?

masseyke: Yeah it does. I haven't figured out how to access that from Java yet though (it's not as straightforward as I'd have thought). I'm currently changing it to use a helper scala class to expose it to java.

jbaiera (approved):
LGTM, thanks for adding this

masseyke:
This is part of #1992

PR #1997, opened 2022-08-30
Ignoring missing indices if es.index.read.missing.as.empty is true

### LINKED ISSUE #1055: Using es.index.read.missing.as.empty returns an empty result if one out of many indices is missing
If a user specifies multiple indices and one of them is not found - if the `es.index.read.missing.as.empty` property is set to true, then the entire connector returns an empty result. Instead, the connector should skip the missing index and continue on with processing the rest of the provided indices.

masseyke:
Fixing the behavior of the `es.index.read.missing.as.empty` setting if one or more indices exist and one or more do not. Previously if any index in the list of indices was missing it would return no results. Now it returns results from indices that do exist and ignores indices that do not exist. Behavior is unchanged if this property is unset or set to false.
Closes #1055


jbaiera (approved):
LGTM!

Issue #2026, opened 2022-09-26
Add an option to pass a truststore in a format of a base64 string


M0arcin:
### What kind an issue is this?

- [ ] Bug report. If you’ve found a bug, please provide a code snippet or test to reproduce it below.   
    The easier it is to track down the bug, the faster it is solved. 
- [x] Feature Request. Start by telling us what problem you’re trying to solve.  
    Often a solution already exists! Don’t send pull requests to implement new features without 
    first getting our support. Sometimes we leave features out on purpose to keep the project small.

### Feature description
The current implementation of passing TrustStore is limited to a path to certain storage - either hdfs or some other location elasticsearch-hadoop could access. It burdens the user with uploading and maintaining the file somewhere, which can be quite a challenge in the case of many connections. 

To simplify the process, Truststore could be loaded from a base64 string, decrypted, and loaded inside the elasticsearch-hadoop package. Of course, such functionality could be allowed on top of the existing solution as an additional parameter.

M0arcin:
PR @ https://github.com/elastic/elasticsearch-hadoop/pull/2027
Not sure why it wasn't connected

PR #2027, opened 2022-09-26
Add an option to pass a truststore in a format of a based64 string (#2026)

M0arcin:

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/commonshttp/SSLSocketFactory.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/commonshttp/SSLSocketFactory.java
@@ -243,7 +243,7 @@ class SSLSocketFactory implements SecureProtocolSocketFactory {
             char[] pass = (StringUtils.hasText(trustStorePass) ? trustStorePass.trim().toCharArray() : null);
             keyStore = loadKeyStoreFromBase64(trustStoreBase64, pass);
         } else {
-            LOG.debug("No truststore nor its specified! SSL is continuing with no truststore.");
+            LOG.debug("No truststore was specified! SSL is continuing with no truststore.");
         }
 
         TrustManagerFactory tmFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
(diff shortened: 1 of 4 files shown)


mr/src/main/java/org/elasticsearch/hadoop/rest/commonshttp/SSLSocketFactory.java
```diff
@@ -224,8 +239,11 @@ private TrustManager[] loadTrustManagers() throws GeneralSecurityException, IOEx
         if (StringUtils.hasText(trustStoreLocation)) {
             char[] pass = (StringUtils.hasText(trustStorePass) ? trustStorePass.trim().toCharArray() : null);
             keyStore = loadKeyStore(trustStoreLocation, pass);
+        } else if (StringUtils.hasText(trustStoreBase64)) {
+            char[] pass = (StringUtils.hasText(trustStorePass) ? trustStorePass.trim().toCharArray() : null);
+            keyStore = loadKeyStoreFromBase64(trustStoreBase64, pass);
         } else {
-            LOG.debug("No truststore location specified! SSL is continuing with no truststore.");
+            LOG.debug("No truststore nor its specified! SSL is continuing with no truststore.");
```
jbaiera, line 246: nit: wording

```suggestion
            LOG.debug("No truststore was specified! SSL is continuing with no truststore.");
```

jbaiera (commented):
I think this looks pretty good. Left one minor nit around wording.

Could you also add some form of test to make sure that SSL auth works with this new base 64 property? Maybe a quick test case added in `org.elasticsearch.hadoop.integration.rest.ssl.SSLTests`?

jbaiera:
For IntelliJ we just import the project by opening the build file and importing as project. Unfortunately a decent amount of the scala project is broken in IntelliJ because they don't support multiple modules using the same code paths which is how the scala cross compiling works in Gradle.

Which imports are missing for you? Are you seeing any specific errors?

jbaiera:
@M0arcin It's been a couple months and we haven't heard back. I'm going to close this for now and if you find time to return to it we can see about reopening.

Issue #1651, opened 2021-04-20
Unable to connect to ElasticSearch server using HTTPS with selfsigned cert


dlubomski:
It looks that lobs try to validate cert. How I can connect without checking this self signet cert ?

Code:

```
reader = spark.read.format("org.elasticsearch.spark.sql") \
    .option("es.port", "9200") \
    .option("spark.es.net.http.auth.user", "login") \
    .option("spark.es.net.http.auth.pass", "password") \
    .option("es.net.proxy.http.host", "https://10.165.92.123") \
    .option("es.net.proxy.http.port", "443") \
    .option("es.net.ssl", "true") \
    .option("es.net.ssl.cert.allow.self.signed", "true") \
    .option("es.nodes.wan.only", "true") \
    .option("es.nodes", "https://10.165.92.146")

df = reader.load("test")
df.registerTempTable('df')
```

Strack trace:

```
Py4JJavaError: An error occurred while calling o1747.load.
: org.elasticsearch.hadoop.EsHadoopIllegalArgumentException: Cannot detect ES version - typically this happens if the network/Elasticsearch cluster is not accessible or when targeting a WAN/Cloud instance without the proper setting 'es.nodes.wan.only'
	at org.elasticsearch.hadoop.rest.InitializationUtils.discoverClusterInfo(InitializationUtils.java:348)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.cfg$lzycompute(DefaultSource.scala:225)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.cfg(DefaultSource.scala:223)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.lazySchema$lzycompute(DefaultSource.scala:229)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.lazySchema(DefaultSource.scala:229)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.$anonfun$schema$1(DefaultSource.scala:233)
	at scala.Option.getOrElse(Option.scala:189)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.schema(DefaultSource.scala:233)
	at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:448)
	at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:326)
	at org.apache.spark.sql.DataFrameReader.$anonfun$load$3(DataFrameReader.scala:308)
[...]

Issue #1146, opened 2018-05-10
Only self-signed certificates of length 1 are trusted


stekiri:
### Issue description

In order to use SSL for Elasticsearch, I set up self-signed certificates using the `certutil` in x-pack. This generates a self-signed certificate that has two entries in the certificate chain. However, the `isTrusted()` method in `SelfSignedStrategy` returns true only if the chain has a length of 1. The connection to Elasticsearch fails as a consequence.

### Steps to reproduce

1. Setting up self-signed certificates following the [instructions](https://www.elastic.co/guide/en/elasticsearch/reference/6.2/configuring-tls.html).
1. Getting info of certificate: `echo "Q" | openssl s_client -connect elastic-url:443 -prexit`
```
---
Certificate chain
 0 s:/CN=instance
   i:/CN=Elastic Certificate Tool Autogenerated CA
 1 s:/CN=Elastic Certificate Tool Autogenerated CA
   i:/CN=Elastic Certificate Tool Autogenerated CA
---
```
3. Writing a dataframe to Elasticsearch:
```
df
  .write
  .mode("append")
  .format("es")
  .option("es.nodes", nodes)
  .option("es.port", port)
  .option("es.index.auto.create", "true")
  .option("es.net.http.auth.user", username)
  .option("es.net.http.auth.pass", password)
  .option("es.resource", index)
  .option("es.nodes.wan.only", "true")
  .option("es.net.ssl.cert.allow.self.signed", "true")
  .option("es.net.ssl", "true")
```

Strack trace:

```
Cannot detect ES version - typically this happens if the network/Elasticsearch cluster is not accessible or when targeting a WAN/Cloud instance without the proper setting 'es.nodes.wan.only'
org.elasticsearch.hadoop.EsHadoopIllegalArgumentException: Cannot detect ES version - typically this happens if the network/Elasticsearch cluster is not accessible or when targeting a WAN/Cloud instance without the proper setting 'es.nodes.wan.only'
	at org.elasticsearch.hadoop.rest.InitializationUtils.discoverEsVersion(InitializationUtils.java:327)
	at org.elasticsearch.spark.sql.EsSparkSQL$.saveToEs(EsSparkSQL.scala:97)
	at org.elasticsearch.spark.sql.ElasticsearchRelation.insert(DefaultSource.scala:610)
	at org.elasticsearch.spark.sql.DefaultSource.createRelation(DefaultSource.scala:103)
[...]

PR #2068, opened 2023-03-08
Fix checking self-signed certs of length greater than 1

j-adamczyk:
This PR fixes issues related to self-signed certificates of length greater than 1: https://github.com/elastic/elasticsearch-hadoop/issues/1651, https://github.com/elastic/elasticsearch-hadoop/issues/1146. The only change required was adopting the check to assert that certificate list is longer than 1, not exactly 1.

This is necessary to work with many certificates. My use case is running Elasticsearch (ECK) and Spark on Kubernetes cluster (https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-tls-certificates.html#k8s-default-self-signed-certificate) and we have self-signed certificates of length 1.

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/rest/commonshttp/SSLSocketFactory.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/rest/commonshttp/SSLSocketFactory.java
@@ -89,7 +89,7 @@ class SSLSocketFactory implements SecureProtocolSocketFactory {
 
     private static class SelfSignedStrategy implements TrustStrategy {
         public boolean isTrusted(X509Certificate[] chain, String authType) throws CertificateException {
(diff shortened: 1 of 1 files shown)


jbaiera:
I don't think this is the right move. A self-signed certificate is by definition a certificate that is issued by itself and signed with its own private key, and therefor should not have more than 1 certificate in the trust chain. Changing the check to allow any number of certificates in the chain is just trusting any certificate that has any number of signers. The setting would no longer be doing what it says it does.

Now, much of the SSL setup process for the Elastic stack uses a getting started process that sets up a self signed CA, but a self-signed CA that signs certificates for your nodes is _not the same_ as a self-signed certificate. If your certs are being signed by something else, even if that something else is self-signed, you should be setting a custom truststore.

PR #2075, opened 2023-03-30
FIxing WritableBytesConverter to not overwrite bytes with wrong value

### LINKED ISSUE #2072: WritableBytesConverter always falls back to JdkBytesConverter

- [X] Bug report. If you’ve found a bug, please provide a code snippet or test to reproduce it below.   
    The easier it is to track down the bug, the faster it is solved. 
- [ ] Feature Request. Start by telling us what problem you’re trying to solve.  
    Often a solution already exists! Don’t send pull requests to implement new features without 
    first getting our support. Sometimes we leave features out on purpose to keep the project small.

### Issue description

WritableBytesConverter always falls back to JdkBytesConverter. I would expect a return after invoking the safeWritableConverter, or a else statement. Previously, this was working correctly, but below commit introduced the behavior of always using the fallback.

https://github.com/elastic/elasticsearch-hadoop/commit/08b2986f5f2759aa9d01a4701cb53b7b6d87a85b

### Steps to reproduce

Code:

Below code is where the unexpected behavior is:
```
@Override
public void convert(Object from, BytesArray to) {

    if (safeWritableConverter != null)
        safeWritableConverter.invoke(from, to);

    super.convert(from, to);
}
```

Strack trace:

N/A

### Version Info

ES-Hadoop   :  7.17.9

masseyke:
This fixes two bugs in `WritableBytesConverter::convert`:

1. If the `from` is a BytesWritable then we first correctly copy its bytes to `to` in `SafeWritableConverter`, and then we incorrectly copy the bytes for the string representation of the byte array into `to` in `JdkBytesConverter` (so a performance problem plus the wrong data in `to`).
2. If the `from` is a Text then we first correctly copy its bytes to `to` in `SafeWritableConverter`, and then we incorrectly copy the string value of `from` into a byte array that might be larger than expected in `JdkBytesConverter` (so a performance problem plus a result array in `to` that is too large).

It looks like this was caused by #937

Closes #2072

### DIFF
--- a/mr/src/main/java/org/elasticsearch/hadoop/mr/SafeWritableConverter.java
+++ b/mr/src/main/java/org/elasticsearch/hadoop/mr/SafeWritableConverter.java
@@ -31,15 +31,26 @@ class SafeWritableConverter {
         Text.class.getName(); // force class to be loaded
     }
 
-    public void invoke(Object from, BytesArray to) {
+    /**
+     * If from is a Text or BytesWritable, the value is written to the to field, and true is returned. Otherwise does not write to
+     * the to field and returns false.
+     * @param from The object to copy bytes from
+     * @param to The BytesArray to copy bytes to
+     * @return true if from has been handled
+     */
+    public boolean invoke(Object from, BytesArray to) {
         // handle common cases
         if (from instanceof Text) {
             Text t = (Text) from;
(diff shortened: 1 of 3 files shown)


jbaiera (approved):
LGTM!

PR #2081, opened 2023-04-21
Updating to spark 3.2.4

masseyke:
Updating spark support from 3.2.3 to 3.2.4 now that it is out.


jbaiera (approved):
LGTM

masseyke:
Unfortunately the build still fails because 3.2.4 is not available in the list of repositories we're using.

PR #2071, opened 2023-03-21
Fix DRA build by removing Gradle 8 deprecation warnings

breskeby:
- Fix dependency info generation on the way

PR #2091, opened 2023-05-03
[8.7] Fix DRA build by removing Gradle 8 deprecation warnings #2071

breskeby:
Fix dependency info generation on the way


mark-vieira (approved):
👍

PR #2092, opened 2023-05-08
Removing Pig, Storm, and Spark 1 support, as well as support for Scala 2.10 on Spark 2

masseyke:
Support for several technologies was deprecated in #1876. This removes support for most of them. The only one mentioned in that PR that is not removed here is Hadoop 1. The reason is that Hive (which is not deprecated or removed) uses old-style mapred input/output formats. So we have to keep it around anyway.


masseyke:
@elasaticmachine run elasticsearch-ci/hadoop-check

masseyke:
Currently blocked by #2099

docs/src/reference/asciidoc/core/intro/download.adoc
```diff
@@ -81,28 +69,14 @@ The Spark connector framework is the most sensitive to version incompatibilities
 | Spark Version | Scala Version | ES-Hadoop Artifact ID
 | 1.0 - 1.2 | 2.10 | <unsupported>
 | 1.0 - 1.2 | 2.11 | <unsupported>
-| 1.3 - 1.6 | 2.10 | elasticsearch-spark-13_2.10
-| 1.3 - 1.6 | 2.11 | elasticsearch-spark-13_2.11
-| 2.0+ | 2.10 | elasticsearch-spark-20_2.10
+| 1.3 - 1.6 | 2.10 | <unsupported> 
+| 1.3 - 1.6 | 2.11 |  <unsupported>
```
jbaiera, line 73: Lets just combine these with the rows above to denote the combined range as unsupported.

jbaiera (approved):
🐖 ⛈️ ➡️ 🥓 🌤️

PR #2099, opened 2023-06-07
Adding a readiness check before using services in tests

masseyke:
We have seen `:qa:kerberos` fail because it tries to run hiveLoadData before the hive server is actually up. We currently only check that the pidfile exists, which really only tells us that we have asked the hive server to start and that it has not crashed. I have seen this happen with hive but it could probably happen with any service if the service for some reason was slow enough coming up. This PR makes it so that each node in each role in each service used gives us a host/port pair, and we check that that port is reachable before attempting to use the node for that role.

### DIFF
--- a/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/services/HadoopServiceDescriptor.groovy
+++ b/buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/services/HadoopServiceDescriptor.groovy
@@ -228,6 +228,48 @@ class HadoopServiceDescriptor implements ServiceDescriptor {
     }
 
     @Override
+    InetSocketAddress readinessCheckHostAndPort(InstanceConfiguration configuration) {
+        RoleDescriptor role = configuration.roleDescriptor
+        SettingsContainer settingsContainer = configuration.getSettingsContainer()
+        if (NAMENODE.equals(role)) {
+            FileSettings hdfsSite = settingsContainer.getFile('hdfs-site.xml')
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/services/HadoopServiceDescriptor.groovy
```diff
@@ -227,6 +227,48 @@ class HadoopServiceDescriptor implements ServiceDescriptor {
...
+    @Override
+    InetSocketAddress readinessCheckHostAndPort(InstanceConfiguration configuration) {
+        RoleDescriptor role = configuration.roleDescriptor
+        SettingsContainer settingsContainer = configuration.getSettingsContainer()
+        if (NAMENODE.equals(role)) {
+            FileSettings hdfsSite = settingsContainer.getFile('hdfs-site.xml')
+            if ('HTTPS_ONLY' == hdfsSite.get('dfs.http.policy')) {
+                return getInetAddress(hdfsSite.getOrDefault('dfs.namenode.https-address', 'localhost:50470'))
+            } else {
+                return getInetAddress(hdfsSite.getOrDefault('dfs.namenode.https-address', 'localhost:50070'))
```
jbaiera, line 239: Should this be `http-address` instead of `https-address`? Looks like this and the datanode else blocks have identical code to the https case.

jbaiera (approved):
LGTM! Right on the money. Left a quick quality check comment, but otherwise looks perfect, thanks!

Issue #1362, opened 2019-10-04
TransportPool validation failures are not handled correctly in Spark environment


nvoskresenskiy:
### What kind an issue is this?

[x ] Bug report.

### Issue description

When elasticsearch-hadoop is used with Spark structured streaming, method org.elasticsearch.hadoop.rest.pooling.TransportPool#validate rethrows a org.elasticsearch.hadoop.EsHadoopException instead of returning false when validation fails 

### Steps to reproduce

1. Create a standard Spark structured streaming query with an Elasticsearch sink. Make sure es http transport pooling is enabled
2. Set up an Elasticsearch cluster with two or more masters
3. Run the query with continuous incoming test data
4. Shut down the first master node

Expected:
Processing continues seamlessly and the transport pool connects to the next master node in the configuration
Actual:
Spark task fails with the stack trace below

Stack trace:

```

19/10/03 21:23:25,628 WARN TaskSetManager: Lost task 2.0 in stage 201.0 (TID 1273, ************, executor 1): org.elasticsearch.hadoop.EsHadoopException: Could not get a Transport from the Transport Pool for host [************:9200]
--
  | at org.elasticsearch.hadoop.rest.pooling.PooledHttpTransportFactory.borrowFrom(PooledHttpTransportFactory.java:110)
  | at org.elasticsearch.hadoop.rest.pooling.PooledHttpTransportFactory.create(PooledHttpTransportFactory.java:56)
  | at org.elasticsearch.hadoop.rest.NetworkClient.selectNextNode(NetworkClient.java:102)
  | at org.elasticsearch.hadoop.rest.NetworkClient.<init>(NetworkClient.java:85)
  | at org.elasticsearch.hadoop.rest.NetworkClient.<init>(NetworkClient.java:61)
  | at org.elasticsearch.hadoop.rest.RestClient.<init>(RestClient.java:99)
  | at org.elasticsearch.hadoop.rest.InitializationUtils.discoverNodesIfNeeded(InitializationUtils.java:93)
  | at org.elasticsearch.hadoop.rest.RestService.createWriter(RestService.java:582)
  | at org.elasticsearch.spark.rdd.EsRDDWriter.write(EsRDDWriter.scala:65)
  | at org.elasticsearch.spark.sql.streaming.EsStreamQueryWriter.run(EsStreamQueryWriter.scala:41)
[...]

PR #2150, opened 2023-10-19
Correctly handling EsHadoopException in TransportPool.validate()

masseyke:
As reported in #1362, when es-hadoop is making requests on behalf of another user, the TransportPool.validate() method incorrectly throws an exception rather than recognizing that the connection is bad and just returning false. The reason is that as of #1244, the exception being thrown in this case is not an IOException but an EsHadoopException with an IOException as its cause. This change checks for that and correctly returns false.


jbaiera (approved):
LGTM

PR #2158, opened 2023-11-09
Always resolve current field name in SparkSQL when creating Row objects inside of arrays

### LINKED ISSUE #2157: Nested objects fail parsing in Spark SQL when empty objects present
We keep track of which field we are currently parsing in the `org.elasticsearch.spark.sql.ScalaRowValueReader#readValue` method:

https://github.com/elastic/elasticsearch-hadoop/blob/4a14860391d00716a5225804a4c71c46a5633162/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala#L39-L46

When reading an array of objects though, the current field that we are reading is overwritten between row creations. We get around this in the create array method by stashing the row order for an array on the call stack:

https://github.com/elastic/elasticsearch-hadoop/blob/4a14860391d00716a5225804a4c71c46a5633162/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala#L76C55-L89

When we create a Row object we check the current field. If we're in an array, we try to use the stashed row order if the current field doesn't have one. If the current field _does_ have one, then we use it instead because we're probably making a subobject under the array:

https://github.com/elastic/elasticsearch-hadoop/blob/4a14860391d00716a5225804a4c71c46a5633162/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala#L60-L69

Unfortunately, if we are parsing a nested document that has an empty document at the very end of it, the empty document field name will remain in the current field variable on the parser. When the next object in the array is created, it will pick up the column list for the previous empty object, which results in downstream serialization issues:

```json
{
  "nested": [          // Current field: `nested`
    {                  // Current field: `nested` (creates map for `nested`)
      "key": "value",  // Current field: `nested.key`
      "object": {}     // Current field: `nested.object` (creates map for `nested.object`)
    },
    {                  // Current field: `nested.object` (creates map for `nested.object` but should have created map for `nested`)
      "key": "value"
    }
  ]
}
```
This isn't a problem if the object has fields because the underlying fields wont have object mappings unless they too are empty objects:

```json
{
  "nested": [             // Current field: `nested`
    {                     // Current field: `nested` (creates map for `nested`)
      "key": "value",     // Current field: `nested.key`
      "object": {         // Current field: `nested.object` (creates map for `nested.object`)
        "subkey": "value" // Current field: `nested.object.subkey`
      }
    },
    {                     // Current field: `nested.object.subkey` (creates map for `nested` using stashed row order because `nested.object.subkey` has no column order data)
      "key": "value"
    }
  ]
}
```

jbaiera:
This PR is meant to fix #2157 

The solution employed is to always resolve the current field name in the value reader when creating Row objects inside of arrays.

Tests were expanded to ensure that no other code was relying on this erroneous behavior. All instances of this happening in the test environment have been cleaned up, further validation put in place, and have had the incorrect behavior confirmed.

### DIFF
--- a/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
+++ b/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
@@ -68,7 +68,10 @@ class ScalaRowValueReader extends ScalaValueReader with RowValueReader with Valu
         // superclass instead of the local sparkRowField.
         var latestRowField = if (getCurrentField == null) null else getCurrentField.getFieldName
         if (latestRowField == null) {
-          latestRowField = Utils.ROOT_LEVEL_NAME
+          throw new IllegalStateException(
+            "No field information could be found while creating map for " +
+              s"array: previous field [${sparkRowField}], row order [${currentArrayRowOrder}]"
+          )
         }
 
         if (rowColumnsMap.contains(latestRowField)) {
--- a/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
+++ b/spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
@@ -59,8 +59,20 @@ class ScalaRowValueReader extends ScalaValueReader with RowValueReader with Valu
     else {
       val rowOrd = 
       if (inArray) {
-        if (rowColumnsMap.contains(sparkRowField)) {
-            rowColumns(sparkRowField)
+        // Recollect the current field name. If the last thing we read before a new object in a list was an empty
+        // object, we won't be able to find the correct row order for the next row being created.
+        // Example: foo:[{bar: baz, qux:{}},{bar:bizzy}]
+        //                              ^   ^____This could break because parser think's that
+        //                               \_________ this field is the current one and loads the wrong row order
+        // By re-resolving the current field, we can avoid this edge case, because that is managed by a stack in the
+        // superclass instead of the local sparkRowField.
+        var latestRowField = if (getCurrentField == null) null else getCurrentField.getFieldName
+        if (latestRowField == null) {
+          latestRowField = Utils.ROOT_LEVEL_NAME
+        }
+
+        if (rowColumnsMap.contains(latestRowField)) {
+            rowColumns(latestRowField)
         }
         else {
           currentArrayRowOrder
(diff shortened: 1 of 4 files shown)


spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
```diff
@@ -59,8 +59,20 @@ class ScalaRowValueReader extends ScalaValueReader with RowValueReader with Valu
...
+        // Recollect the current field name. If the last thing we read before a new object in a list was an empty
+        // object, we won't be able to find the correct row order for the next row being created.
+        // Example: foo:[{bar: baz, qux:{}},{bar:bizzy}]
+        //                              ^   ^____This could break because parser think's that
+        //                               \_________ this field is the current one and loads the wrong row order
+        // By re-resolving the current field, we can avoid this edge case, because that is managed by a stack in the
+        // superclass instead of the local sparkRowField.
+        var latestRowField = if (getCurrentField == null) null else getCurrentField.getFieldName
+        if (latestRowField == null) {
+          latestRowField = Utils.ROOT_LEVEL_NAME
```
masseyke, line 71: I'm curious when we would ever hit this, and I'm not 100% certain this is what we'd want if we did. Maybe just an assertion that we're not here?

masseyke (approved):
I had one question, but it looks good to me!

spark/sql-30/src/main/scala/org/elasticsearch/spark/sql/ScalaEsRowValueReader.scala
jbaiera: I don't think landing in this code is possible, but I wanted to be more lenient if we ever did. But you know what they say about leniency...

I've updated the conditional to throw an exception with debugging info.

jbaiera:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.11|[](https://github.com/elastic/elasticsearch-hadoop/pull/2166)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

jbaiera:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|7.17|[](https://github.com/elastic/elasticsearch-hadoop/pull/2167)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #2237, opened 2024-06-06
Fix Spring and conjars for 6.8

zdunecki:
It took much time since last update on older version of `elasticsearch-hadoop` and build stops working. In my company we use Elasticsearch `6.8.x` that's why it would be helpful for us to have this fix in this version. 

I see some version has same problems but currently I'm focused to fix that in version where our dependency is.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -8,6 +8,8 @@ apply plugin: 'groovy'
 
 boolean localRepo = project.getProperties().containsKey("localRepo")
 
+println localRepo
+
 Properties props = new Properties()
 props.load(project.file('esh-version.properties').newDataInputStream())
 version = props.getProperty('eshadoop')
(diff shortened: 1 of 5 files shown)


masseyke:
Hi @zdunecki. Could you describe the problem you are solving? It looks like the intention is for this to go into a 6.x es-hadoop release. Unfortunately 6.x is long past end of life and won't get any more releases.

zdunecki:
@masseyke 
I see. Generally speaking for this version (and not only) the build is not possible due the changes of spring and conjars repository url. It's just a fix for make build successfully. 

Since at my company we depdens on Elsticsearch 6.8 and I fix this internally (and it works) I thought it may be helpful for community. 

If you do not accept changes anymore then you can close this PR but otherwise tell my how can I make this production ready due to conflitcs and CLA (but I accepted that yesterday).

masseyke:
I think what you are doing is probably your best option. Es-hadoop is released in sync with Elasticsearch, and there will almost certainly never be another 6.8.x release (the last bugfix release was more than 2 years ago).

zdunecki:
Sure thing. 

Leaving the history of this PR for the community may be helpful. If there will be another 6.8.x release then you can think about that once again.

Currently, I'm closing this PR.

PR #2305, opened 2024-12-24
Deprecate Spark 2.x

samxbr:
Apache Spark 2.x has reached end of support since 2021, we are deprecating Spark 2.x in 8.18. Support for it will be removed in 9.0. This change will be backported to 8.x branch.

PR #2316, opened 2025-01-03
Remove Spark 2.x

samxbr:
After Spark 2.x deprecation in 8.18 https://github.com/elastic/elasticsearch-hadoop/pull/2305, this change removes support for it in 9.0.


docs/src/reference/asciidoc/appendix/breaking.adoc
```diff
@@ -8,6 +8,16 @@ For clarity, we always list any breaking changes at the top of the
...
 
+[[breaking-changes-90]]
+=== Breaking Changes in 9.0
+
+This section details the breaking changes when upgrading {eh} from 8.x to 9.0.
+
+[[removals-9.0]]
+==== Removal of Spark 2.x
+
+Support for the Spark 2.x has been removed in {eh} 9.0, Spark 3.x is still supported.
```
jbaiera, line 19: Lets instead say that Spark 3.x is the new default support version

spark/core/build.gradle
```diff
@@ -10,12 +10,10 @@ apply plugin: 'spark.variants'
 sparkVariants {
     capabilityGroup 'org.elasticsearch.spark.variant'
 
-    // Changing the formatting of these lines could break .buildkite/pipeline.py, it uses regex to parse the `spark20scala212` part
+    // Changing the formatting of these lines could break .buildkite/pipeline.py, it uses regex to parse the `spark30scala212` part
```
jbaiera, line 13: We should check in with delivery about this line to see if anything needs to be updated in CI for this change

jbaiera (approved):
LGTM pending comments addressed. I think we'll need to make sure that we are logging a deprecation warning when Spark 2.x is used in 8.x. We may also want to discuss bumping the default scala supported version to 2.13 instead of leaving it on 2.12 considering how long these major releases live for, but that can be a follow up after more discussion.

spark/core/build.gradle
samxbr: Sure, I also found this [old PR of yours](https://github.com/elastic/infra/pull/39243) that removes the old Spark from release manager.

samxbr:
> LGTM pending comments addressed. I think we'll need to make sure that we are logging a deprecation warning when Spark 2.x is used in 8.x. We may also want to discuss bumping the default scala supported version to 2.13 instead of leaving it on 2.12 considering how long these major releases live for, but that can be a follow up after more discussion.

+1 on making Scala 2.13 the default, I can try doing that as a follow up.

samxbr:
> LGTM pending comments addressed. I think we'll need to make sure that we are logging a deprecation warning when Spark 2.x is used in 8.x. We may also want to discuss bumping the default scala supported version to 2.13 instead of leaving it on 2.12 considering how long these major releases live for, but that can be a follow up after more discussion.

I added deprecation warning in https://github.com/elastic/elasticsearch-hadoop/pull/2305 and backported to 8.x, do you think we want to add more?

PR #2328, opened 2025-01-21
Set Spark 3.x default to Scala 2.13

samxbr:
This change upgrades the default Scala version from 2.12 to 2.13 for Spark 3.x artifacts.


buildSrc/src/main/groovy/org/elasticsearch/hadoop/gradle/fixture/hadoop/services/SparkYarnServiceDescriptor.groovy
```diff
@@ -78,7 +78,8 @@ class SparkYarnServiceDescriptor implements ServiceDescriptor {
     @Override
     String artifactName(ServiceConfiguration configuration) {
         Version version = configuration.getVersion()
-        return "spark-$version-bin-hadoop${hadoopVersionCompatibility()}"
+        // artifact name: https://spark.apache.org/downloads.html
+        return "spark-$version-bin-hadoop${hadoopVersionCompatibility()}-scala2.13"
```
samxbr, line 82: Thanks @masseyke for finding this.

masseyke:
Just so I understand -- the impact here is that now our uber jar (which I would advise people not not use anyway) will use scala 2.13, and more of our testing (specifically maybe just the kerberos testing?) will use scala 2.13. Right? Or is there another impact that I'm missing?

masseyke (approved):
LGTM

PR #2349, opened 2025-02-26
[docs] Migrate docs from AsciiDoc to Markdown

colleenmcginnis:
Migrate docs from AsciiDoc to Markdown. The preview can be built after #2348  is merged.

@lcawl please tag reviewers, add the appropriate label(s), and take this out of draft when you're ready.

Note: I'm not sure if it was intentional, but it looks like the toc structure is different than the AsciiDoc version. 👀


lcawl:
> Note: I'm not sure if it was intentional, but it looks like the toc structure is different than the AsciiDoc version. 👀 

I don't think it was intentional, so I've done significant edits to the content and navigation in this PR. Thanks for the callout!

lcawl (approved):
LGTM

PR #2513, opened 2026-04-02
Adding support for Spark 3.5, 4.0, and 4.1 

masseyke:
### Summary

  - Adds `elasticsearch-spark-35`, `elasticsearch-spark-40`, and `elasticsearch-spark-41` artifacts to support Spark 3.5, 4.0, and 4.1 alongside the existing Spark 3.4 artifact
  - Consolidates the heavily duplicated Spark SQL source into a single shared directory (`spark/sql-shared/`) 
  - Each Spark version runs its integration tests on the Java version it requires (Spark 3.x on Java 8, Spark 4.0 on Java 17, Spark 4.1 on Java 21)

  ### Spark API differences handled per version

  | Version | Breaking change |
  |---|---|
  | Spark 3.5 | `RowEncoder` API changed to `RowEncoder.encoderFor(schema, lenient=false)` wrapped in `ExpressionEncoder`; `spark-common-utils` and `spark-sql-api` split out as separate artifacts |
  | Spark 4.0 | `Dataset.sqlContext` removed — replaced with `Dataset.sparkSession`; Java 17 required |
  | Spark 4.1 | `MetadataLog` removed from Spark internals entirely — streaming commit protocol reimplemented using a local trait and direct Hadoop FS writes; Java 21 required |

  ### Source sharing approach

  All source files that are identical across Spark versions live in `spark/sql-shared/src/` and are
  pulled into each module via Gradle `srcDir`. Only the six main-source files and two itest files that
  genuinely differ between versions remain per-module. The four build artifacts (`elasticsearch-spark-30`,
  `-35`, `-40`, `-41`) and their test suites are otherwise unchanged.

PR #2539, opened 2026-07-06
Consolidate shared Spark SQL sources into sql-shared

masseyke:
## Summary

Pure refactor extracted from #2513 to keep that PR focused on new-version support.

Moves the Spark SQL/streaming sources that are identical across Spark versions out of `spark/sql-30` into a new `spark/sql-shared/` directory, and wires them back into `sql-30` via `srcDir` (main, test, and itest source sets).

- **34 files move byte-for-byte** (GitHub should show them as renames).
- The only content edit is `spark/sql-30/build.gradle`, which adds the shared source directories.
- **No functional change.**

## Why

Prepares for adding `elasticsearch-spark-35/40/41`, which will each reference the same shared sources instead of duplicating them.


ankurs47 (approved):
LGTM. Build is failing. probably not related to the change.

masseyke:
I just checked the [build on main](https://buildkite.com/elastic/elasticsearch-hadoop-tests/builds?branch=main), and it has been failing with this same error for several days. So I don't think it is caused by this PR.

PR #2544, opened 2026-07-09
Add support for Spark 3.5

ankurs47:
Adds the `elasticsearch-spark-35` artifact to support Spark 3.5

Spark 3.5 itself is low-risk: it still runs on Java 8, reuses `spark/sql-shared/` (#2539)
for all common code, and only **two source files** differ from `sql-30`.

## Spark 3.5 API differences handled

| Change | Handling |
|---|---|
| `RowEncoder` API changed | Use `RowEncoder.encoderFor(schema, lenient = false)` wrapped in `ExpressionEncoder`, in `EsStreamQueryWriter.scala` |
| `spark-common-utils` / `spark-sql-api` split out as separate artifacts | Spark 3.5 moved `org.apache.spark.internal.Logging`, `Row`, `StructType`, etc. out of `spark-core`/`spark-sql`; added as explicit dependencies in `spark/core/build.gradle` and `spark/sql-35/build.gradle` |
| `StreamingQueryListener` fires idle events | Spark 3.5+ fires `onQueryIdle` instead of a 0-row progress event when there's no new data; handled in the shared `StreamingQueryTestHarness.scala` test harness |


spark/sql-35/build.gradle
```diff
@@ -0,0 +1,215 @@
...
+
+apply plugin: 'java-library'
+apply plugin: 'scala'
+apply plugin: 'es.hadoop.build.integration'
+apply plugin: 'spark.variants'
+
+
+sparkVariants {
+    capabilityGroup 'org.elasticsearch.spark35.sql.variant'
+    // Spark 3.5+ only supports Scala 2.13 (spark-common-utils/spark-sql-api are not published for Scala 2.12)
```
masseyke: It turns out that this isn't true -- spark 3.5 also supports scala 2.12. I am not certain that we need to support 2.12 on spark 3.5 though. @EdLewisEL do you know if we need to support scala 2.12 on spark 3.5? It's doable, but it adds yet another artifact we need to build and test.

masseyke:
@ankurs47 now that we have spark-common, can we get rid of all the duplicate `src` classes? I think we can, and only keep `RowEncoder` and `ExpressionEncoder` (the only two that differ). I haven't tried it though, so maybe there's some reason we need to keep the others? I don't thinks so though.

spark/core/build.gradle
```diff
@@ -14,6 +14,7 @@ sparkVariants {
     // We should maybe move these to a separate config file that can be read from both this file and the pipeline script in the future if it creates issues
     setCoreDefaultVariant "spark30scala213", spark30Version, scala213Version
     addCoreFeatureVariant "spark30scala212", spark30Version, scala212Version
+    addCoreFeatureVariant "spark35scala213", spark35Version, scala213Version
```
masseyke: This line causes `./gradlew dist` to fail with:
```
   > Publishing is not able to resolve a dependency on a project with multiple publications that have different coordinates.
     Found the following publications in project ':elasticsearch-spark':
       - Maven publication 'spark30scala212' with coordinates org.elasticsearch:elasticsearch-spark_2.12:9.6.0-SNAPSHOT
       - Maven publication 'spark35scala213' with coordinates org.elasticsearch:elasticsearch-spark_2.13:9.6.0-SNAPSHOT
```

spark/sql-35/build.gradle
masseyke: Alternatively, we could correct this comment and change the PR description to "Add support for Spark 3.5 on Scala 2.13", and potentially do 2.12 later if anyones requests it.

masseyke:
We'll need to update the compatibility matrix at https://github.com/elastic/elasticsearch-hadoop/blob/main/docs/reference/installation.md. I think that is the only documentation that needs updating, but I haven't done a thorough review.

dist/build.gradle
```diff
@@ -48,14 +48,18 @@ configurations {
 
 BuildPlugin.disableTransitiveDependencies(project, project.configurations.thirdPartyShaded)
 
-def distProjects = [":elasticsearch-hadoop-mr", ":elasticsearch-hadoop-hive", ":elasticsearch-spark-30"]
+def distProjects = [":elasticsearch-hadoop-mr", ":elasticsearch-hadoop-hive", ":elasticsearch-spark-30", ":elasticsearch-spark-35"]
```
masseyke: I think handling it like this is going to cause problems. This is used to build up the hadoop uber jar. I think it sort of works because it rejects duplicates and so spark-30 wins out. But it's kind of dangerous. (Side note: I think putting support for a random spark/scala distribution in the hadoop uber jar only causes problems anyway, and I'd like to remove it altogher, but that's out of scope for this PR).

spark/core/build.gradle
masseyke: This one is actually fairly critical b/c CI does not catch it, so we'd have a clean build. Then it would fail when we tried to release. Ideally, we'd add something to our standard CI run to catch this, but that's probably out of scope for this PR.

ankurs47:
> We'll need to update the compatibility matrix at https://github.com/elastic/elasticsearch-hadoop/blob/main/docs/reference/installation.md. I think that is the only documentation that needs updating, but I haven't done a thorough review.

I added the docs.

ankurs47:
> @ankurs47 now that we have spark-common, can we get rid of all the duplicate `src` classes? I think we can, and only keep `RowEncoder` and `ExpressionEncoder` (the only two that differ). I haven't tried it though, so maybe there's some reason we need to keep the others? I don't thinks so though.

I did some refactoring. Now most of the code is in shared. Let me know if this makes sense.

masseyke:
It looks like with the latest changes, none of the spark-35 integration tests (on scala 2.12 or 2.13) are executed by CI. Or at least, they're not explicitly labeled like the spark-30 tests (I'm looking under buldkite/elasticsearch-hadoop-tests). I think they might still be in the intake step, but they probably ought to be pulled out into their own steps that can run in parallel, like the spark-30 ones.

ankurs47:
> It looks like with the latest changes, none of the spark-35 integration tests (on scala 2.12 or 2.13) are executed by CI. Or at least, they're not explicitly labeled like the spark-30 tests (I'm looking under buldkite/elasticsearch-hadoop-tests). I think they might still be in the intake step, but they probably ought to be pulled out into their own steps that can run in parallel, like the spark-30 ones.

Just pushed a fix to the pipeline for this. I will monitor and see if it is doing the correct thing.

masseyke (approved):
LGTM

PR #2546, opened 2026-07-17
Add support for Spark 4.0

ankurs47:
## Summary

- Adds `spark/sql-40` module (Scala 2.13 only; Java 17 minimum per Spark 4.0 requirements)
- `EsRowEncoder` uses Spark 4.0's new `ExpressionEncoder.apply(StructType, boolean)` factory
- Tests and javadoc use Java 17+: honors `JAVA17_HOME`, falls back to lowest discovered JVM ≥ 17 (on CI this resolves to `openjdk21` via `JAVA_HOME`)
- Skips genjavadoc — it generates `abstract static` stubs that Java 17 javadoc rejects
- `--add-opens` workaround for securemock's CGLIB on Java 9+ (comment in build file explains root cause and long-term fix)


spark/sql-40/src/main/scala/org/elasticsearch/spark/sql/streaming/EsRowEncoder.scala
```diff
@@ -0,0 +1,10 @@
+package org.elasticsearch.spark.sql.streaming
```
masseyke, line 20: I think we need the standard license header here.

ankurs47: Added the license.

spark/sql-shared/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
```diff
@@ -1117,6 +1117,10 @@ class AbstractScalaEsScalaSparkSQL(prefix: String, readMetadata: jl.Boolean, pus
 
   @Test
   def testDataSourcePushDown08InWithNumberAndStrings() {
+    // Spark 4+ defaults spark.sql.ansi.enabled=true, which rejects implicit casts of 'bar'/'foo'
+    // to BIGINT and throws SparkNumberFormatException instead of returning 0 results.
+    // The ANSI rejection is strictly correct behaviour, so skip rather than fail.
+    assumeTrue(!org.apache.spark.SPARK_VERSION.startsWith("4"))
```
masseyke, line 1123: So that we don't have to revisit this particular line when we add support for spark 5 or 6, maybe this ought to be `assumeTrue(org.apache.spark.SPARK_VERSION.startsWith("3"))`?

masseyke (approved):
Looks really good to me

spark/sql-shared/src/itest/scala/org/elasticsearch/spark/integration/AbstractScalaEsSparkSQL.scala
ankurs47: Sounds good. I will merge this for now as is and change it in 4.1 pr.

